PR #6, opened 2015-07-24
KEY-87 - Add metrics endpoint

jacobhaven:
Add metrics endpoint using go-metrics to measure the number of requests and their average response latency on the keyserver.

Running metrics HTTP server can be seen here: http://keyserver.jhaven.me:2408/metrics?indent=1

A standard output should look something like:
```
{
        "Invalid Request Rate": {
                "15m.rate": 0,
                "1m.rate": 0,
                "5m.rate": 0,
                "count": 0,
                "mean.rate": 0
        },
        "Request Rate": {
                "15m.rate": 0.002179432534806779,
                "1m.rate": 0.02490778899904623,
                "5m.rate": 0.006288975787638508,
                "count": 2,
                "mean.rate": 0.050002368910979464
        },
        "Response Latency": {
                "15m.rate": 0.002179432534806779,
                "1m.rate": 0.02490778899904623,
                "5m.rate": 0.006288975787638508,
                "75%": 7.38846e+06,
                "95%": 7.38846e+06,
                "99%": 7.38846e+06,
                "99.9%": 7.38846e+06,
                "count": 2,
                "max": 7388460,
                "mean": 3.799837e+06,
                "mean.rate": 0.050002421981064166,
                "median": 3.799837e+06,
                "min": 211214,
                "stddev": 3.588623e+06
        }
}
```

### DIFF
--- a/server/metrics.go
+++ b/server/metrics.go
@@ -1,0 +1,81 @@
+package server
+
+import (
+	"bytes"
+	"encoding/json"
+	"log"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/cloudflare/go-metrics"
+)
+
+type statistics struct {
+	rate        metrics.Meter
+	invalidRate metrics.Meter
+	latency     metrics.Timer
+	metrics.Registry
+}
+
+func newStatistics(metricsAddr string) *statistics {
+	stats := &statistics{
+		rate:        metrics.NewMeter(),
+		invalidRate: metrics.NewMeter(),
+		latency:     metrics.NewTimer(),
+		Registry:    metrics.NewRegistry(),
+	}
+	stats.Register("Request Rate", stats.rate)
+	stats.Register("Invalid Request Rate", stats.invalidRate)
+	stats.Register("Response Latency", stats.latency)
+
+	go stats.ListenAndServe(metricsAddr)
+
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -214,18 +228,25 @@ func (s *Server) handle(conn *gokeyless.Conn) {
...
 
 		connError = conn.Respond(h.ID, sig)
+		s.stats.logRequest(requestBegin)
 	}
+
 	s.Log.Printf("Connection error: %v\n", connError)
-	return
+}
+
+func (s *Server) serveMetrics() {
```
bobrik: why is it empty?

server/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/cloudflare/go-metrics"
+)
+
+// // totalRequests is the total number of requests received.
+// 	totalRequests int
+// 	// validRequests is the total number of requests responded to.
```
bobrik: this might be aligned

server/server.go
jacobhaven: Both good points. Because I didn't clean them up when I separated out the two files. *fixed*

PR #7, opened 2015-07-28
Updates to synchronization, connection handling, and other bug fixes

jacobhaven:
Various bug fixes found using keyless tester.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -5,11 +5,13 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
 	"math/rand"
 	"net"
+	"sync"
 
 	"github.com/cloudflare/gokeyless"
 )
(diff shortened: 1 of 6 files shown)


server/server.go
```diff
@@ -177,6 +177,8 @@ func (s *Server) handle(conn *gokeyless.Conn) {
 		case gokeyless.OpECDSASignSHA1:
 			opts = crypto.SHA1
 		case gokeyless.OpRSASignSHA224:
+			isRSA = true
```
lziest: can we group the condition in which isRSA is assigned true? the sequence of alternative 'fallthrough's makes it a little hard to maintain in my opinion.

have second switch on Opcode {
case gokeyless.OpRSASignSHA224, gokeyless.OpRSASignSHA256, gokeyless.OpRSASignSHA384, gokeyless.OpRSASignSHA512:
isRSA=true
}

and did opts assignment like
case gokeyless.OpRSASignSHA224, gokeyless.OpECDSASignSHA224:
opts = crypto.SHA224
case gokeyless.OpRSASignSHA256, gokeyless.OpECDSASignSHA256:
...

jacobhaven: Not a bad idea (I had actually forgotten this was valid syntax). This use actually eliminates the need for the `isRSA` flag altogether.

client/client.go
```diff
@@ -79,26 +89,34 @@ func (c *Client) Dial(server string) (*gokeyless.Conn, error) {
...
-			existing = append(existing, conn)
+		c.m.RLock()
+		conn, ok := c.conns[server]
+		c.m.RUnlock()
+		if ok {
+			if conn.Use() {
+				return conn, nil
+			}
+			c.m.Lock()
+			if c.conns[server] == conn {
```
jacobhaven: This could be a loop that breaks when `c.conns[server] == conn` (and `!conn.Use()`)

PR #12, opened 2015-08-08
Update GetDigest interface to match GetSKI

jacobhaven:
Now returns `ok bool` instead of `error`.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -47,6 +47,7 @@ type defaultKeystore struct {
 }
 
 // Add adds a new key to the server's internal repertoire.
+// Stores in maps by SKI and (if possible) Digest, SNI, Server IP, and Client IP.
 func (keys *defaultKeystore) Add(op *gokeyless.Operation, priv crypto.Signer) error {
 	ski, err := gokeyless.GetSKI(priv.Public())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jacobhaven:
@lziest Fair. I really just care about them being consistent, so I swapped GetDigest to return `"can't comute digest for non-RSA public key"` instead of `false` and reset GetSKI. 
I'll rebase before merging if this looks good.

server/server.go
```diff
@@ -56,7 +56,7 @@ func (keys *defaultKeystore) Add(op *gokeyless.Operation, priv crypto.Signer) er
 	keys.Lock()
 	defer keys.Unlock()
 
-	if digest, ok := gokeyless.GetDigest(priv.Public()); ok {
+	if digest, err := gokeyless.GetDigest(priv.Public()); err == nil {
```
lziest: return the err if it's not nil?

jacobhaven: No, if GetDigest will only succeed for RSA keys. We should just discard the error it throws for ECDSA keys (and not have a digest registered for that key).

lziest: Thanks! put it in the comment?

jacobhaven: It's sort of implicit, just the same as all the following optional map keys. But, sure.

jacobhaven:
*rebased and squashed commits* Looks good to merge?

lziest:
LGTM

PR #17, opened 2015-08-28
Updates to gokeyless-tester

jacobhaven:
Much refactoring to allow use of shared code in `tests` or `client` libraries.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -4,13 +4,18 @@ import (
 	"crypto"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
 	"net"
+	"os"
+	"path/filepath"
+	"regexp"
 	"sync"
 
+	"github.com/cloudflare/cfssl/helpers"
 	"github.com/cloudflare/cfssl/log"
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -172,3 +177,93 @@ func (c *Client) RegisterPublicKey(server string, pub crypto.PublicKey) (*Privat
 func (c *Client) RegisterCert(server string, cert *x509.Certificate) (*PrivateKey, error) {
 	return c.RegisterPublicKey(server, cert.PublicKey)
 }
+
+// RegisterDir reads all .pubkey and .crt files from a directory and returns associated PublicKey structs.
+func (c *Client) RegisterDir(server, dir string, LoadPubKey func([]byte) (crypto.PublicKey, error)) (privkeys []*PrivateKey, err error) {
+	var pubkeyExt, crtExt *regexp.Regexp
+	if pubkeyExt, err = regexp.Compile(`.+\.pubkey`); err != nil {
```
kisom: Maybe these regexps should be moved to global `MustCompile`s.

jacobhaven: Seems reasonable. \**adds `var` block right above the function*\*

kisom:
LGTM.

PR #25, opened 2015-10-02
Add SNI and ServerIP to tester input

jacobhaven:
This is needed to test lazy-loading with the `gokeyless-tester`

### DIFF
--- a/client/keys.go
+++ b/client/keys.go
@@ -13,11 +13,9 @@ import (
 
 // PrivateKey represents a keyless-backed RSA private key.
 type PrivateKey struct {
-	public crypto.PublicKey
-	ski    gokeyless.SKI
-	digest gokeyless.Digest
-
-	client *Client
+	public   crypto.PublicKey
+	client   *Client
+	template *gokeyless.Operation
 
 	crypto.Signer
 	crypto.Decrypter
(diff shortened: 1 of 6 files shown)


client/keys.go
```diff
@@ -13,11 +13,9 @@ import (
...
 // PrivateKey represents a keyless-backed RSA private key.
 type PrivateKey struct {
-	public crypto.PublicKey
-	ski    gokeyless.SKI
-	digest gokeyless.Digest
-
-	client *Client
+	public   crypto.PublicKey
+	client   *Client
+	template *gokeyless.Operation
```
lziest: do you think "opTemplate" would be better? Naming is hard, and  English is my 2nd language. :cry:

jacobhaven: Maybe, It's just so long

kisom: You also see `tpl` a lot. I'd be fine with that.

jacobhaven: I'm really just fine with `template`. Any other comments?

lziest: do you like to create a new Metadata struct here? and call it md? Decouple from gokeyless.Operation? key.template just confuses me a little bit.

jacobhaven: That's fair. It was mainly in order to be extensible with other/new `gokeyless.Operation` info.

lziest:
LGTM

PR #26, opened 2015-10-08
Add support for blacklisting server IPs in the client

jacobhaven:


jacobhaven:
This is intended to provide a way to prevent a gokeyless server/client from recursively dialing itself by preventing registering anything that could resolve to the same IP address as one of the server's SANs.

lziest:
LGTM

PR #29, opened 2015-11-11
Support for latency-based load balancing across multiple key servers

jacobhaven:
Initial stab at latency based load balancing across multiple server IPs using a priority queue of server Addrs.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -12,7 +12,6 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"strconv"
 	"sync"
 
 	"github.com/cloudflare/cfssl/helpers"
@@ -28,10 +27,8 @@ type Client struct {
 	Dialer *net.Dialer
 	// DefaultRemote is a default remote to dial and register keys to.
 	DefaultRemote Remote
-	// BlacklistIPs is a list of server IPs that this client won't dial.
-	BlacklistIPs []net.IP
-	// BlacklistPort is the server port to blacklist.
-	BlacklistPort int
+	// Blacklist is a list of addresses that this client won't dial.
+	Blacklist AddrSet
 	// m is a Read/Write lock to protect against conccurrent accesses to maps.
 	m sync.RWMutex
 	// servers maps all known server names to their servers.
@@ -51,9 +48,10 @@ func NewClient(cert tls.Certificate, keyserverCA *x509.CertPool) *Client {
 				tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
 			},
 		},
-		Dialer:  &net.Dialer{},
-		servers: make(map[string]Remote),
-		remotes: make(map[gokeyless.SKI]Remote),
+		Dialer:    &net.Dialer{},
+		Blacklist: make(AddrSet),
+		servers:   make(map[string]Remote),
+		remotes:   make(map[gokeyless.SKI]Remote),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


client/remote.go
```diff
@@ -0,0 +1,134 @@
...
+			if err := conn.Ping(nil); err != nil {
+				log.Infof("Connection to %s failed: %v", s, err)
+				return
+			}
+			duration := time.Since(start)
+
+			r.Lock()
+			s.latency.Update(duration)
+			if s.index >= 0 {
+				heap.Fix(r, s.index)
```
lziest: would this cause out-of-bound error? s is not longer in r?

jacobhaven: It should be in `r` (it's only removed removed while the lock is helf), but if it isn't `s.index` is set to `-1` (thus the `>=0` check)

lziest: got it!

lziest:
CI error is git error :(
LGTM.

client/remote.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (sh serverHeap) Swap(i, j int) {
+	sh[i], sh[j] = sh[j], sh[i]
+	sh[i].index = i
+	sh[j].index = j
+}
+
+func (sh serverHeap) Less(i, j int) bool {
+	return sh[j].conn == nil || sh[i].latency.RateMean() < sh[j].latency.RateMean()
```
grittygrease: Is RateMean the right metric to be using? Recent requests should be more highly weighted than old requests.

client/client.go
```diff
@@ -116,26 +81,10 @@ func (c *Client) DialAny(ski gokeyless.SKI) (*gokeyless.Conn, error) {
...
-			}
-			c.m.Lock()
-			if c.conns[server] == conn {
-				delete(c.conns, server)
-			}
-			c.m.Unlock()
-		}
-	}
-
 	// choose from possible servers at random until a connection can be established.
```
jacobhaven: This is a bit of a punt. Really there should only be one `Remote` registered. We could enforce that by having `registerSKI` append to the current remote (creating a `group` if necessary).

client/client.go
```diff
@@ -197,31 +170,75 @@ func (c *Client) ValidServer(server string) (err error) {
...
-// registerSKI associates the SKI of a public key with a particular keyserver.
-func (c *Client) registerSKI(server string, ski gokeyless.SKI) error {
-	if server == "" {
-		server = c.DefaultServer
+// AddRemote adds a remote for the given SKI, optionally registering it
+// under the given server name.
+func (c *Client) AddRemote(server string, r Remote, ski gokeyless.SKI) *PrivateKey {
+	c.m.Lock()
+	defer c.m.Unlock()
+
```
jacobhaven: Make sure this logic all makes the most sense.

client/client.go
```diff
@@ -77,151 +75,125 @@ func NewClientFromFile(certFile, keyFile, caFile string) (*Client, error) {
...
+// AddRemote adds a remote for the given SKI, optionally registering it
+// under the given server name.
+func (c *Client) AddRemote(server string, r Remote, ski gokeyless.SKI) *PrivateKey {
+	c.m.Lock()
+	defer c.m.Unlock()
 
-	var ips []net.IP
-	if ips, err = net.LookupIP(host); err != nil {
-		return
+	if server != "" {
```
jacobhaven: I could use some eyes on all the cached `c.servers` and `c.remotes` logic to make sure it make sense.

lziest:
lgtm again!

PR #31, opened 2015-12-01
Add AKI Whitelisting ability to defaultKeystore

jacobhaven:


jacobhaven:
This won't come up in the normal read-keys-from-disk usage of the keystore, as the `Operation` for those keys is just set to `nil`.

kisom:
LGTM.

jkroll-cf:
LGTM.

PR #32, opened 2016-01-18
Bugfix: enforce SKI match in keystore lookup if provided in op

jacobhaven:
This was causing issues when multiple certs had the same SNI, but not the same SKI (i.e. certificate fallback)


lziest:
lgtm

tlianza:
Do we have any unit tests for this codepath? Seems like this could be an easy behavior to accidentally revert.

PR #33, opened 2016-01-20
Custom DNS Resolver Support

jacobhaven:


lziest:
lgtm

FiloSottile:
Super late, but LGTM. Will work with any sane resolver.

jacobhaven:
@FiloSottile In production, we're seeing `dns: failed to unpack truncated message` What's the best way to handle large DNS record lists?

FiloSottile:
Ah, right. So, over IPv4 you can use very high (4k) EDNS0 sizes (add an OPT record to the query, there's example code in RRDNS's vdns filter).

Over IPv6 you have to stay in the MTU, which limits you to about 1200, which might not be enough, in which case you are stuck with seeing the Truncate bit and retrying over TCP.

PR #37, opened 2016-02-09
Update keyserver initialization

jacobhaven:
TODO: ensure server cert isn't expired and automatically renew with `ActivationToken` if it is.

TODO: Set token.json in pkg/


kisom:
LGTM.

PR #39, opened 2016-02-24
Replace go-metrics with prometheus

jacobhaven:


server/metrics.go
```diff
@@ -1,33 +1,37 @@
...
-		Registry:    metrics.NewRegistry(),
+		requestsTotal: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: "requests_total",
+			Help: "Number of total requests.",
+		}),
+		requestsInvalid: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: "requests_invalid",
+			Help: "Number of invalid requests.",
+		}),
+		latency: prometheus.NewGauge(prometheus.GaugeOpts{
```
dqminh: latency should be better represented with Histogram or Summary.

jacobhaven: Agreed. Updated to use a `Summary`.

dqminh:
LGTM

PR #40, opened 2016-02-29
Initialization/Packaging updates

jacobhaven:


pkg/gokeyless.sysv
```diff
@@ -61,28 +53,14 @@ start() {
...
-      mkdir `$dirname SERVER_CERT` || return 1
-    fi
-
-    $program -init -key="${SERVER_KEY}" -cert="${SERVER_CERT}"
-
-  fi
-
   [ -f $pidfile ] && rm -f $pidfile
 
+  cd $WORKDIR
```
jacobhaven: We could check here if `token.json` exists, and then create it if it doesn't. I mainly didn't want to marshal JSON from Bash.

lziest: I wouldn't either

pkg/gokeyless.sysv
```diff
@@ -61,28 +53,14 @@ start() {
...
   sh -c "
     exec \"$program\" $args
-  " >>/var/log/$name.log 2>>/var/log/$name.err &
-
-  wait $!
+  " 2>>/var/log/$name.log &
 
-  for i in 0 1 2 3 4 5; do
+  for i in 0 1 2 3 4 5 6 7 8 9 10; do
```
lziest: more retry?

jacobhaven: If the keyserver doesn't have a cert, it has to make an API call out to certificateAPI to get one, which then makes requests back to the keyserver for 5s. Only when it's properly initialized will it write the PID file.

lziest:
lgtm

PR #43, opened 2016-03-10
Handle non-JSON response bodies.

elithrar:
For non-200 responses we now attempt to unmarshal the body as JSON. If that fails, we return the HTTP status code for troubleshooting purposes.

(We could, in addition, log the response headers, but I'm attempting to avoid cluttering the user-facing logs)

### DIFF
--- a/cmd/gokeyless/initialize.go
+++ b/cmd/gokeyless/initialize.go
@@ -71,19 +71,27 @@ func initAPICall(token *apiToken, csr string) ([]byte, error) {
 	}
 	defer resp.Body.Close()
 
-	apiResp := new(initAPIResponse)
-	if err := json.NewDecoder(resp.Body).Decode(apiResp); err != nil {
-		return nil, err
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("certificate API returned an invalid response body for HTTP %d", resp.StatusCode)
 	}
+
+	apiResp := &initAPIResponse{}
+	if err := json.Unmarshal(bodyBytes, apiResp); err != nil {
+		log.Debugf("invalid JSON response: %s", bodyBytes)
+		return nil, fmt.Errorf("certificate API returned HTTP %d", resp.StatusCode)
+	}
+
 	if !apiResp.Success {
 		errs, _ := json.Marshal(apiResp.Errors)
-		return nil, fmt.Errorf("api call failed: %s", errs)
+		return nil, fmt.Errorf("certificate API call failed: %s", errs)
 	}
 
 	if cert, ok := apiResp.Result["certificate"]; ok {
 		return []byte(cert), nil
 	}
-	return nil, fmt.Errorf("no certificate in api response: %#v", apiResp)
+
+	return nil, fmt.Errorf("no certificate in API response: %#v", apiResp)
 }
 
 func getToken() (*apiToken, error) {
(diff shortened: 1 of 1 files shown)


jacobhaven:
LGTM

cmd/gokeyless/initialize.go
```diff
@@ -71,19 +71,29 @@ func initAPICall(token *apiToken, csr string) ([]byte, error) {
 	}
 	defer resp.Body.Close()
 
-	apiResp := new(initAPIResponse)
+	apiResp := &initAPIResponse{}
+
+	if resp.StatusCode != http.StatusOK {
+		// Check if the returned body is JSON by attempting to unmarshal. If it
+		// fails, return the HTTP status code for debugging.
+		apiErr, err := json.Marshal(apiResp.Errors)
```
kisom: The comment says unmarshal, but the code says `Marshal`. What am I missing?

cmd/gokeyless/initialize.go
```diff
@@ -71,19 +71,29 @@ func initAPICall(token *apiToken, csr string) ([]byte, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("API returned HTTP %d", resp.StatusCode)
```
kisom: Or, `gokeyless: API...`

cmd/gokeyless/initialize.go
jacobhaven: This should `ioutil.ReadAll(resp.Body)` to a `body []byte`, then attempt to Unmarshal it to `apiResp`. If it fails, it should log the raw body and `resp.StatusCode`. Otherwise
```Go
if !apiResp.Success {		
	errs, _ := json.Marshal(apiResp.Errors)
	return nil, fmt.Errorf("api call failed: %s", errs)
}
```

elithrar: Fixing.

cmd/gokeyless/initialize.go
jacobhaven: Also good.

kisom:
Once the noted fixes are in, LGTM.

elithrar:
Updated.

cmd/gokeyless/initialize.go
```diff
@@ -71,19 +71,35 @@ func initAPICall(token *apiToken, csr string) ([]byte, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("certificate API returned an invalid response body for HTTP %d", resp.StatusCode)
+	}
+
+	apiResp := &initAPIResponse{}
+	if resp.StatusCode != http.StatusOK {
+		// Check if the returned body is JSON by attempting to unmarshal it into
+		// our response struct. If it fails, return the HTTP status code and raw
+		// body for debugging.
+		err := json.Unmarshal(bodyBytes, apiResp)
```
jacobhaven: Get rid of this `resp.StatusCode != http.StatusOK` check and `err := json.NewDecoder(resp.Body).Decode(apiResp)` below. You've already read the `resp.Body` into `bodyBytes`, so it can't be passed to the decoder later.

Just call 

```
if err := json.Unmarshal(bodyBytes, apiResp); err != nil {
	log.Debugf("invalid JSON response: %s", bodyBytes)
	return nil, fmt.Errorf("certificate API returned HTTP %d", resp.StatusCode)
}
if !apiResp.Success {		
 	errs, _ := json.Marshal(apiResp.Errors)		
 	return nil, fmt.Errorf("api call failed: %s", errs)		
 }
```

elithrar:
@kisom Re-review on Monday please.

kisom:
LGTM.

PR #51, opened 2016-04-07
Modify the header used for initialization auth

jkroll-cf:
To be consistent with the CloudFlare API, use the right header for service tokens, necessary for https://api.cloudflare.com/client/v4/certificates/.


kisom:
It looks like `go vet` is no longer part of the `golang.org/x/tools` package; so you can remove the `go get` from the `.travis.yml`.

kisom:
If https://github.com/cloudflare/gokeyless/pull/52 works, we can merge that and rebase this PR.

kisom:
LGTM.

PR #53, opened 2016-04-12
Use a more sensible name for the service API key.

jkroll-cf:
This name is consistent with the API documentation.


lziest:
LGTM, need to remove 'go vet' in Travis scripts later.

elithrar:
👍

PR #54, opened 2016-04-13
[ci] Refactored .travis.yml

elithrar:
- Removed unused deps
- Updated script to ignore vendor dir correctly
- Run the race detector for tests
- Implicitly email recipients on success AND failure.

This should also fix the go vet issue.


kisom:
LGTM.

PR #55, opened 2016-04-13
[cmd/gokeyless] Make manual activation more robust.

elithrar:
- Now writes the CSR to disk
- Quits if an existing CSR (default: server.csr) is found to avoid generating a
  new key and invalidating any "in flight" CSRs.
- Requires manual removal (`rm server.csr`) to reset process.

### DIFF
--- a/cmd/gokeyless/gokeyless.go
+++ b/cmd/gokeyless/gokeyless.go
@@ -15,7 +15,10 @@ import (
 	"github.com/cloudflare/gokeyless/server"
 )
 
-var defaultEndpoint = "https://api.cloudflare.com/client/v4/certificates/"
+var (
+	defaultEndpoint = "https://api.cloudflare.com/client/v4/certificates/"
+	csrFile         = "server.csr"
+)
 
 var (
 	initToken    string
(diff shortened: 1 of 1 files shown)


elithrar:
Note: although `-manual-activation` is a fallback, making it robust/providing a better UX is still useful.

cmd/gokeyless/gokeyless.go
```diff
@@ -108,3 +116,24 @@ func LoadKey(in []byte) (priv crypto.Signer, err error) {
...
+	}
+
+	f, err := os.Create(csrFile)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	_, err = f.Write(csr)
+	if err != nil {
```
jkroll-cf: This could be one-lined with the `f.Write` above.

kisom:
Aside from @jkroll-cf's comments, LGTM.

cmd/gokeyless/gokeyless.go
elithrar: Fixed.

elithrar:
Updated.

jkroll-cf:
LGTM.

Issue #56, opened 2016-04-15
Race condition in `listenResponse`


kisom:
[One of the Travis tests](https://travis-ci.org/cloudflare/gokeyless/jobs/122697769) found a data race in [`listenResponse`](https://github.com/cloudflare/gokeyless/blob/master/conn.go#L127).

```
WARNING: DATA RACE
Write by goroutine 23:
  runtime.mapdelete()
      /home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/hashmap.go:545 +0x0
  github.com/cloudflare/gokeyless.(*Conn).listenResponse.func1()
      /home/travis/gopath/src/github.com/cloudflare/gokeyless/conn.go:130 +0xf4
  github.com/cloudflare/gokeyless.(*Conn).listenResponse()
      /home/travis/gopath/src/github.com/cloudflare/gokeyless/conn.go:138 +0x258
  github.com/cloudflare/gokeyless.(*Conn).DoOperation()
      /home/travis/gopath/src/github.com/cloudflare/gokeyless/conn.go:149 +0xc5
  github.com/cloudflare/gokeyless/client.(*PrivateKey).execute()
      /home/travis/gopath/src/github.com/cloudflare/gokeyless/client/keys.go:91 +0x403
  github.com/cloudflare/gokeyless/client.(*PrivateKey).Decrypt()
      /home/travis/gopath/src/github.com/cloudflare/gokeyless/client/keys.go:130 +0x234
  github.com/cloudflare/gokeyless/tests.TestRSADecrypt()
      /home/travis/gopath/src/github.com/cloudflare/gokeyless/tests/client_test.go:116 +0x52d
  testing.tRunner()
      /home/travis/.gimme/versions/go1.6.linux.amd64/src/testing/testing.go:473 +0xdc
Previous read by goroutine 24:
  runtime.mapaccess1_fast32()
      /home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/hashmap_fast.go:13 +0x0
  github.com/cloudflare/gokeyless.(*Conn).listenResponse()
      /home/travis/gopath/src/github.com/cloudflare/gokeyless/conn.go:138 +0x1f5
  github.com/cloudflare/gokeyless.(*Conn).DoOperation()
      /home/travis/gopath/src/github.com/cloudflare/gokeyless/conn.go:149 +0xc5
  github.com/cloudflare/gokeyless.(*Conn).Ping()
[...]

PR #57, opened 2016-04-20
Don't defer channel shutdown.

kisom:
This was causing a data race (e.g. #56). Instead of deferring this, it should be done explicitly.


jkroll-cf:
Causes test failures?

kisom:
Those test failures are what this was trying to fix. (The failures only seem to show up in Travis.) Closing because it didn't work.

PR #61, opened 2016-05-11
Copy TLS config explicitly, to avoid copying a mutex

### LINKED ISSUE #49: Duplicate tls.Config without copying sync.Once
```
$ go vet $(go list ./... | grep -v /vendor/)
client/remote.go:118: assignment copies lock value to config: crypto/tls.Config contains sync.Once contains sync.Mutex
```

jkroll-cf:
This fixes a go vet issue.

Fixes #49

### DIFF
--- a/client/remote.go
+++ b/client/remote.go
@@ -115,10 +115,10 @@ func (s *server) Dial(c *Client) (*gokeyless.Conn, error) {
 		return s.conn, nil
 	}
 
-	config := *c.Config
+	config := copyTLSConfig(c.Config)
 	config.ServerName = s.ServerName
 	log.Debugf("Dialing %s at %s\n", s.ServerName, s.String())
-	inner, err := tls.DialWithDialer(c.Dialer, s.Network(), s.String(), &config)
+	inner, err := tls.DialWithDialer(c.Dialer, s.Network(), s.String(), config)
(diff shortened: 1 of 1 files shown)


client/remote.go
```diff
@@ -132,6 +132,17 @@ func (s *server) Add(r Remote) Remote {
 	return g
 }
 
+func copyTLSConfig(c *tls.Config) *tls.Config {
```
elithrar: Would it make more sense (e.g. be more robust) to copy the entire structure, in the event other fields come into play? (however unlikely)

jkroll-cf: Fair enough. I've added that. Kind of a mess, though.

elithrar:
```
func copyTLSConfig(a *tls.Config) *tls.Config {
    b := &tls.Config{}
   *b = *a
    return b
}
```

jacobhaven: Why not [reflect.Copy()](https://golang.org/pkg/reflect/#Copy)

elithrar: @jacobhaven Need a copy without copying the mutex (private field), otherwise you can just do `*b = *a` to copy the datastructure underneath.

lziest:
LGTM

PR #63, opened 2016-05-27
Zi/bug fixes

lziest:


lziest:
a lot of data race with connections, I need to dive into this.

conn.go
```diff
@@ -127,7 +127,6 @@ func (c *Conn) listenResponse(id uint32) (*Header, error) {
 	defer func() {
 		c.Lock()
 		close(c.listeners[id])
-		delete(c.listeners, id)
```
lziest: @kisom @elithrar that seems to fix the race condition. But...do you like it?

kisom: I'm not sure --- it means we're not cleaning up closed listeners.

kisom: It's probably better to make sure that map accesses are protected, or look at doing this a different way.

lziest: I added it back, and get around the data race by not using map[key] over and over again, since we already do value, ok := map[key], we can just use the value, instead of reading map.

lziest:
I think I get around the race. 😆

kisom:
That was the approach I was thinking of.

kisom:
LGTM.

lziest:
edited out a typo in one commit message.

lziest:
merging

PR #64, opened 2016-06-07
remove dead codes

lziest:


jkroll-cf:
LGTM.

tests/tests.go
```diff
@@ -35,11 +35,6 @@ func NewPingTest(c *client.Client, server string) testapi.TestFunc {
 			return err
 		}
 
-		cookie := make([]byte, 512)
```
jacobhaven: `cookie` should actually be passed into `Ping` rather than `nil` (to test that the remote server actually echoes the data)

PR #65, opened 2016-06-09
move flag definition in cfssl/log's init() into main's init()

lziest:


kisom:
LGTM.

lziest:
This is a good start, but vendoring makes managing 3rd party package exported variable super confusing. Alice vendors log, Bob vendors Alice and log, but B can only control B's log behaviors only, Alice's log package is still out of control...

lziest:
I feel like the history is repeating itself: shared lib vs archived lib

jkroll-cf:
It seems odd to me to make changes to a vendored package, especially when we control the upstream.

lziest:
@jkroll-cf it's a result of vendor update.

lziest:
but it is confusing since I thought this update may help me but it actually doesn't. It's rooted in the vendoring itself. Essentially doing nested vendoring needs to be very very very careful.

PR #66, opened 2016-06-10
Zi/make vendor right

lziest:
1. Remove root vendor and move it to cmd/, create a symbolic link for gokeyless itself under cmd/vendor/github.com/cloudflare/gokeyless
2. fix another data race so CI is green.


lziest:
@FiloSottile please review.

Makefile
```diff
@@ -33,8 +33,12 @@ install-config:
 	@install -m400 pkg/testing-ecdsa.key $(CONFIG_PREFIX)/keys/testing-ecdsa.key
 	@install -m400 pkg/testing-rsa.key $(CONFIG_PREFIX)/keys/testing-rsa.key
 
-$(INSTALL_BIN)/$(NAME): | install-config
-	@GOOS=linux GOARCH=$(ARCH) go build -o $@ ./cmd/$(NAME)/...
+.PHONY: get-gvt
+get-gvt:
+	go get -u github.com/FiloSottile/gvt
```
jkroll-cf: Trying to promote a world where we git clone && git checkout -f <revision>, since `go get` doesn't have much of an integrity story.

FiloSottile: I agree with the effort 100%, but without aggressive fsck settings, git offers no integrity against malicious adversaries.

If the goal is just to pin the version, I think that's not what that make target wants.


Sent from a small keyboard

lziest:
this is kinda urgent

cmd/gokeyless-cli-tester/vendor/github.com/cloudflare/gokeyless
```diff
@@ -0,0 +1 @@
+../../../../../../gokeyless
```
FiloSottile: This should be `../../../../..` in case the folder is not called gokeyless.

FiloSottile:
I'm generally not a fan of `gvt restore` as it breaks `go get`, but your call. If you keep it, I'd change https://github.com/cloudflare/gokeyless#source-installation to mandate the Makefile though.

`fix another data race on hashmap` seems unrelated.

FiloSottile:
To avoid triple bloat, you can put `vendor/` in `cmd/`.

lziest:
I tried `vendor/` under `cmd/` but it doesn't work with `go build cmd/gokeyless` then, I haven't tried `go build cmd/...` though

FiloSottile:
That's strange, it's really supposed to work the same however command you use to build.

Try just moving the vendor folder up one level and then running the exact same "go build".

lziest:
It's probably I didn't do the path symlink right. Now it is working with cmd/vendor

lziest:
And the error message was complaining about gokeyless/client.go can't find dependency packages

lziest:
@kisom @FiloSottile @jkroll-cf review it again?

jkroll-cf:
Not being super clear on vendor rules or best practices, if @FiloSottile is happy and the code builds, I'm happy. In the longer future, we should work to isolate the functionality we need to use as a library into packages that we can treat as libraries everywhere, which will moot the underlying problem.

LGTM, for the moment.

lziest:
I confirm that when I vendor and import github.com/clouflare/gokeyless, the log package being picked up by gokeyless is now locally 'go get' or vendored cfssl/log, not the nested vendored one.

PR #67, opened 2016-06-14
introduce Conn and a unexported connection pool into client

lziest:
1. remove the reference count in gokeyless.Conn. The use of such
   reference count is questionable since our goal is simply keeping
   the underlying TLS connection alive.

2. Add a Conn type into client, which wraps around gokeyless.Conn.
   This Conn type associates the original gokeyless.Conn with a
   server address and a health checking goroutine.

   When initializing client.Conn, a recurrent goroutine is spawned
   to monitor the connection.

3. Keep all established client.Conn in a hashmap based connection pool,
   indexed by remote address. When a connection is not longer usable,
   either discovered by the corresponding health checker or the keyless
   operation, the connection is removed from the pool. When connecting
   to a remote address, always check with connection pool first.

4. Using client to dial a remote address will return
   client.Conn, rather than gokeyless.Conn. And so we can make a
   client-based gokeyless proxy with automatic health check and
   connection management. Through testing, we can see that when
   a remote server shuts down, the corrsponding connection will be
   closed, and new keyless request will trigger new connection
   attempts, rather than keeping using the broken one.

### DIFF
--- a/client/keys.go
+++ b/client/keys.go
@@ -79,9 +79,8 @@ func (key *PrivateKey) execute(op gokeyless.Op, msg []byte) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer conn.Close()
 
-	result, err := conn.DoOperation(&gokeyless.Operation{
+	result, err := conn.Conn.DoOperation(&gokeyless.Operation{
 		Opcode:   op,
 		Payload:  msg,
 		SKI:      key.ski,
@@ -91,6 +90,7 @@ func (key *PrivateKey) execute(op gokeyless.Op, msg []byte) ([]byte, error) {
 		SNI:      key.sni,
 	})
 	if err != nil {
+		defer conn.Close()
 		return nil, err
 	}
 
--- a/client/remote.go
+++ b/client/remote.go
@@ -16,17 +16,113 @@ import (
 	"github.com/miekg/dns"
 )
 
+// connPoolType is a async safe pool of established gokeyless Conn
+// so we don't need to do TLS handshake unnecessarily.
+type connPoolType struct {
+	pool map[string]*Conn
+	sync.RWMutex
+}
+
+// connPool keeps all active Conn
+var connPool connPoolType
+
 // A Remote represents some number of remote keyless server(s)
 type Remote interface {
-	Dial(*Client) (*gokeyless.Conn, error)
+	Dial(*Client) (*Conn, error)
 	Add(Remote) Remote
+}
+
+// A Conn represents a long-lived client connection to a keyserver.
+type Conn struct {
+	*gokeyless.Conn
+	addr string
+	ch   chan bool // a channel that will be read by health-check goroutine
 }
 
 // A singleRemote is an individual remote server
 type singleRemote struct {
-	net.Addr
-	ServerName string
-	conn       *gokeyless.Conn
+	net.Addr          // actual address
+	ServerName string // hostname for TLS verification
+}
+
+func init() {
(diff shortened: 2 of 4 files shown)


lziest:
@elithrar @kisom @jkroll-cf @FiloSottile @Bren2010 @grittygrease  please review.

lziest:
@jacobhaven please review when you have time. :)

client/remote.go
```diff
@@ -16,17 +16,113 @@ import (
...
-	Dial(*Client) (*gokeyless.Conn, error)
+	Dial(*Client) (*Conn, error)
 	Add(Remote) Remote
 }
 
+// A Conn represents a long-lived client connection to a keyserver.
+type Conn struct {
+	*gokeyless.Conn
+	addr string
+	ch   chan bool // a channel that will be read by health-check goroutine
```
jacobhaven: `ch` is a bit nondescript. Maybe `healthCh`?

lziest:
fixing data race...

client/remote.go
lziest: sure

client/remote.go
```diff
@@ -16,17 +16,113 @@ import (
...
+				// start a new timer
+				go healthchecker(conn, ch, after)
+				ch <- true
+			}
+		}
+	}
+}
+
+// Get returns a Conn from the pool if there is any.
+func (p connPoolType) Get(key string) *Conn {
```
jacobhaven: I think you want these to be pointer receivers. Otherwise I think you're copying the mutex and doing all sorts of bad things.

client/remote.go
```diff
@@ -16,17 +16,113 @@ import (
...
+				err := conn.Ping(nil)
+				if err != nil {
+					log.Debug("health check ping failed:", err)
+					// shut down the conn
+					conn.Close()
+					return
+				}
+				log.Debug("start a new health check timer")
+				// start a new timer
+				go healthchecker(conn, ch, after)
```
jacobhaven: Wouldn't this be easier with a loop instead of spawning a recursive goroutine?

lziest:
I have tested this gokeyless client change using a nginx+keyserver proxy+keyserver setup. It works fine and the performance is on par with san-proxy setup.

lziest:
Based on testing, I will merge this PR for now since it is blocking other projects.

client/keys.go
```diff
@@ -91,6 +90,7 @@ func (key *PrivateKey) execute(op gokeyless.Op, msg []byte) ([]byte, error) {
 		SNI:      key.sni,
 	})
 	if err != nil {
+		defer conn.Close()
```
elithrar: Does this need to be a `defer` here since you are returning anyway? Is the conn being kept open when no error has occurred?

lziest: yes it can be closed immediately but I just copy it from below. And yes the conn is kept open when no error has occurred. For each connection, we have a separate thread which keeps using the connection for ping. If the connection goes bad, it is closed.

client/remote.go
```diff
@@ -16,17 +16,103 @@ import (
...
+		}
+	}
+}
+
+// Get returns a Conn from the pool if there is any.
+func (p *connPoolType) Get(key string) *Conn {
+	p.RLock()
+	defer p.RUnlock()
+
+	conn, ok := p.pool[key]
```
kisom: If ok is false, conn will be nil since it's the default value; you could just do 

```
conn := p.pool[key]
return conn
```

Assigning to a var and then returning the var should make sure that the map access occurs before the call to `Unlock`.

PR #70, opened 2016-06-20
add a test on connection selection with optimized performance

lziest:
1. refactor client tests a little bit
2. add a test that sets up two backend servers. One server is running
   with artifical slow connection, and the other is unchanged. Test
   that client can pick the connection with better speed.
3. refactored server code a little bit such that NewServer() won't register
   global prometheus metrics. So we can run two servers side by side in
   tests otherwise prometheus will complain about duplicate metrics
   registration.

### DIFF
--- a/client/remote_test.go
+++ b/client/remote_test.go
@@ -5,7 +5,10 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"io/ioutil"
+	"net"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/cloudflare/gokeyless"
 	"github.com/cloudflare/gokeyless/server"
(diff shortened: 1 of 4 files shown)


client/remote_test.go
```diff
@@ -104,55 +105,105 @@ func TestMain(t *testing.T) {
...
+	_, err := c.Dial(ecdsaSKI)
+	if err == nil {
+		t.Fatal("bad remote management")
 	}
+}
 
-	if pemBytes, err = ioutil.ReadFile(ecdsaPubKey); err != nil {
+func TestSlowServer(t *testing.T) {
+	// Setup keyless server
+	s2, err := server.NewServerFromFile(serverCert, serverKey, keylessCA,
```
elithrar: Could we bind to `0` instead of binding to a static port? Avoids potential conflicts when testing in a shared env.

jkroll-cf:
LGTM, modulo @elithrar's comment, which I agree with (I've had this sort of pain on CI systems before - it's not fun).

kisom:
Agree with @elithrar, otherwise LGTM.

client/remote_test.go
lziest: agree!

elithrar:
LGTM.

PR #75, opened 2016-06-29
client: return a pointer to RSAPrivateKey

FiloSottile:


lziest:
lgtm, I have a test that can be improved later.

PR #79, opened 2016-07-01
client package test tuning: sleep 100ms so server starts cleanly

lziest:


lziest:
this reduces the possibility to hit the race condition on client dialing "localhost:0"

kisom:
LGTM.

PR #81, opened 2016-07-06
Zi/new tests

lziest:
Add supports to benchmark testing and certificate load request testing.


tests/tests.go
```diff
@@ -61,6 +61,31 @@ func NewActivateTest(c *client.Client, server string, hashedToken []byte) testap
 	}
 }
 
+// NewPingTest generates a TestFunc to connect and perform a certificate load.
```
lziest: need to fixup comments

PR #84, opened 2016-07-15
move backoff timer into client.Conn, removing workload variables and …

lziest:
…eliminating r/w race condition


lziest:
never mind, close this, the race condition has very minor side effect though.

PR #87, opened 2016-07-26
SKI string format should be hex encoding

lziest:
base64 encoding doesn't make sense


kisom:
LGTM, pending tests.

lziest:
let me update comment.

PR #88, opened 2016-07-26
SKI string format should be hex encoding

lziest:
base64 encoding doesn't make sense


kisom:
Good catch on the comment. Still LGTM, pending tests.

PR #90, opened 2016-07-27
Add support for RSASSA-PSS signatures

FiloSottile:
@lziest @grittygrease


protocol.go
```diff
@@ -19,6 +19,8 @@ import (
 	"github.com/cloudflare/cfssl/helpers"
 )
 
+//go:generate stringer -type=Tag,Op -output=protocol_string.go
```
lziest: do I need to be extra careful with 'go generate' inside the vendored code?
In other words, can we checked in the ugly generated protocol_string.go so I don't need to worry about it?

lziest: nvm, github hides it by default, now I see it.

lziest:
lgtm

PR #92, opened 2016-08-10
RegisterPublicKey should return crypto.Signer and does return *RSAPri…

lziest:
…vateKey when necessary


lziest:
@grittygrease please review, This way we don't need to cast PrivateKey to RSAPrivateKey at various places.

Bren2010:
I don't understand this change. The point of RSAPrivateKey is to conditionally implement the Decrypt method.

PR #100, opened 2016-08-18
refactor blacklisting and ip lookups

lziest:
- export blacklisting API with individual host
- make LookupIPs package util function


kisom:
LGTM.

PR #104, opened 2016-08-19
simpler SKI remote update mechanism

lziest:
- eliminate Add function in Remote interface
- eliminate hashmap as Remote lookup cache


lziest:
travis won't run this branch, close this one and re-open under a new branch name

PR #108, opened 2016-08-22
add wait for test server to start in proxy_test.go

lziest:


kisom:
LGTM, but might be useful to see [how RO does this](https://github.com/cloudflare/redoctober/blob/master/redoctober_test.go#L103).

PR #110, opened 2016-08-24
Allow explicit PSS salt size == hash size

FiloSottile:


client/keys.go
```diff
@@ -18,7 +18,8 @@ func signOpFromSignerOpts(key *PrivateKey, opts crypto.SignerOpts) gokeyless.Op
 		if _, ok := key.Public().(*rsa.PublicKey); !ok {
 			return gokeyless.OpError
 		}
-		if opts.SaltLength != rsa.PSSSaltLengthEqualsHash {
+		if opts.SaltLength != rsa.PSSSaltLengthEqualsHash &&
```
lziest: please comment inline to explain why it is 'and'?

So this is allowed, opts.SaltLength != rsa.PSSSaltLengthEqualsHash && opts.SaltLength == opts.Hash.Size(), because?

FiloSottile: Added a comment

lziest:
lgtm

PR #111, opened 2016-08-26
Support limited concurrency on the wire.

Bren2010:
- The client-side locked itself into being serial. Fix that.
- Make the server's read-loop delegate to a pool of workers.
- Also, stop pinging all the time.

### DIFF
--- a/client/remote.go
+++ b/client/remote.go
@@ -317,8 +317,7 @@ type item struct {
 // A Group is a Remote consisting of a load-balanced set of external servers.
 type Group struct {
 	sync.Mutex
-	remotes     []*item
-	lastPingAll time.Time
+	remotes []*item
 }
 
 // NewGroup creates a new group from a set of remotes.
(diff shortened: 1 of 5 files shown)


lziest:
It will be nice to have concurrency tests. I have been thinking about this for a while.

lziest:
travis-ci is not acting correctly recently, probably because of OAuth token being invalid?

Bren2010:
I don't feel like the build failures are my fault?

Please let me know what you think of the test and also, don't miss the "LockOSThread" part. I don't know if LockOSThread should be configureable or something; it makes the first request really slow while enough threads are being acquired, but it also makes performance *much* more consistent, as the OS has dumber context-switching logic.

lziest:
I probably need to study a little bit about LockOSThread.

client/remote.go
```diff
@@ -369,38 +370,43 @@ func (g *Group) Dial(c *Client) (conn *Conn, err error) {
...
-				i.latency.Reset()
-				i.errorCount++
-				log.Infof("ping failed: %v", err)
-			} else {
-				log.Debug(conn.addr, " ping duration:", duration)
-				i.latency.Update(duration)
-			}
+func (g *Group) pingAll(c *Client) {
+	time.Sleep(100 * time.Microsecond)
+	g.Lock()
```
lziest: I actually want to rewrite this part as well. Do you mind reviewing my changes soonish?
The "pingAll" mechanism shouldn't lock the group in practical uses. Because a group can have hundreds of remotes, pinging each of them sequentially takes a lot of time, factoring in timeout/error etc. So I decide to make latency measurement less aggressive and more opportunistic.

Bren2010: I have no problem with that. I'll watch my notification box.

lziest: I have rewritten this part, can you left client/remote.go untouched in your PR? so we won't have merge conflicts.

lziest:
conn and server changes, lgtm.

lziest:
regarding the flaky error, does "SIGQUIT: quit" make sense to you?

lziest:
I think it is my fault, the connection pool cache has deadlocks? LGTM

PR #113, opened 2016-08-29
remove Add() method from Keystore interface

lziest:
- export DefaultKeystore
- LoadKeysFromDirectory is detached from Server


cmd/gokeyless/gokeyless.go
```diff
@@ -101,9 +103,11 @@ func main() {
 		select {
 		case <-c:
 			log.Info("Received SIGHUP, reloading keys...")
-			if err := s.LoadKeysFromDir(keyDir, LoadKey); err != nil {
+			keys := server.NewDefaultKeystore()
```
lziest: This also fixes a bug that existing keys never get purged from the memory. Now they will be gc'ed.

kisom:
LGTM.

PR #114, opened 2016-08-29
simplify DefaultKeyserver

lziest:
- hashmaps that persists the relation between SNI, IPs with SKI are removed
  we don't need those relations persisted forever in the memory. But we don't
  have mechanism to erase those relations. So those maps are considered
  problematic.
- remove AKI whitelisting check for now until we decide what to do with it
  with a design documentation.


lziest:
build failure

lziest:
builds are now passing. I need to figure out travis ci's permission right now

Bren2010:
LGTM

kisom:
LGTM.

Issue #115, opened 2016-08-30
Group Dial's latency measurement is problematic


lziest:
With mutex lock, a remote Group can either do dial and return connection, or do health and latency measurement. Since latency measurement would possibly take very long to finish, it practically prevent the remote group to carry out dial immediately. 

We should make Dial to return as fast as possible, while measurement check can be carried out with low priority.

PR #116, opened 2016-08-30
address #115, Group Dial should never wait for Group.pingAll

lziest:
- simplify the Group structure
- latency now also tracks error counts, eliminating needs on unexport
  type 'item'
- stop sorting or heapify the internal slice, which ask for mutex lock
  on update, and we don't need to implement interfaces. Less code, less
  bugs.

### DIFF
--- a/client/remote.go
+++ b/client/remote.go
@@ -1,12 +1,10 @@
 package client
 
 import (
-	"container/heap"
 	"crypto/tls"
 	"errors"
 	"fmt"
 	"net"
-	"sort"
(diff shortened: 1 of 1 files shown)


lziest:
@Bren2010 now please review.

lziest:
duh, the first Dial of a Group remote needs to be a pingAll...localhost resolving to both ipv4/ipv6 is indeed good tests.

lziest:
I am closing this since I rewrite it for a 2nd time.

PR #117, opened 2016-08-31
address #115, rewrite Group Dial

lziest:
- it satisfy three design goals
  - load balancing through shuffle top n remotes (n=3, adjustable)
  - robust failover: it will try all top n remotes when dial failure happens
  - dynamic server reordering based on ping performances by a separate
    goroutine and proper locking.

- the turnaround time of Dial function is kept low.
- the background ping measurement in PingAll can take arbitrary long time
  without affecting Dial.

### DIFF
--- a/client/remote.go
+++ b/client/remote.go
@@ -149,6 +149,10 @@ func UnixRemote(unixAddr, serverName string) (Remote, error) {
 	return NewServer(addr, serverName), nil
 }
 
+// LookupIPs resolves host with resolvers list sequentially unitl
+// one resolver can answer the request. It falls
+// back to use system default for final resolution if none of resolvers
+// can answer.
 func LookupIPs(resolvers []string, host string) (ips []net.IP, err error) {
 	m := new(dns.Msg)
 	dnsClient := new(dns.Client)
--- a/client/remote.go
+++ b/client/remote.go
@@ -1,10 +1,10 @@
 package client
 
 import (
-	"container/heap"
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"math/rand"
 	"net"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


lziest:
@Bren2010 please review this.

client/remote.go
```diff
@@ -326,121 +327,119 @@ func NewGroup(remotes []Remote) (*Group, error) {
...
+	}()
 
-	for _, f := range popped {
-		heap.Push(g, f)
+	// shuffle remotes for load balancing
+	for i := range remotes {
+		j := rand.Intn(i + 1)
+		if i != j {
+			remotes[i], remotes[j] = remotes[j], remotes[i]
+		}
```
Bren2010: If you're trying to do Fisher-Yates, this seems backwards?
Maybe instead of copying and doing an in-place shuffle, you could do [Fisher-Yates inside-out](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_.22inside-out.22_algorithm).

lziest: I want to operate on a copy so we don't change the original copy because it is a temporary, for load balancing purpose only.

Bren2010: I believe I understand what you're trying to do. The reason I proposed the inside-out algorithm is because it does the copying and shuffling in one step.

Also, I don't believe what you've implemented is Fisher-Yates, because the index is going from 0 to n-1, rather than n-1 to 0. The inside-out algorithm doesn't have this problem.

lziest: got it, I will fix.

lziest: On a 2nd thought, I think I did inside-out with extra copying.

Bren2010:
LGTM

lziest:
squashed and rebased.

PR #118, opened 2016-08-31
remove connectiom pool eviction func for now

lziest:
- nested locking needs some more thoughts


Bren2010:
LGTM pending tests

lziest:
@Bren2010 I simply want to avoid nested lock

Issue #112, opened 2016-08-28
document the usage of AKI whitelist


lziest:
currently, we have AKI field of a gokeyless operation.
https://github.com/cloudflare/gokeyless/blob/master/protocol.go#L285
which is used to whitelist gokeyless operation in keyserver.
https://github.com/cloudflare/gokeyless/blob/master/server/server.go#L134

We need to write the design of this feature, because it seems it is not being used and could cause confusion and major bugs.
The default keyserver would not populate AKI whitelist. see https://github.com/cloudflare/gokeyless/blob/master/server/server.go#L401
because op == nil, AKI whitelist is skipped.
https://github.com/cloudflare/gokeyless/blob/master/server/server.go#L72

lziest:
Let's just remove AKI, as it appears really hard to handle whitelist updates. Its append-only interface is not well thought about. And I have a PR to remove it.

PR #119, opened 2016-09-02
address #112, remove AKI whitelist

lziest:
The AKI whitelist has apparent usage of allowing certain keys only to
answer operations from a certain keyless protocol client,
whose tls certificate issuer is on the list.

There are problems with current hashmap based implementation.

1. Currently whitelist is Add-only, which makes it impractical for
   whitelist CRUD.
2. Keyserver can have multiple keystores for each authorized keyless
   client. Each keystore has the agility of dynamic
   key loading. We can simply ask a corresponding keystore mapped by
   client cert issuer id: keyserver has the final say of whether a
   key is present for a given client.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -99,10 +99,6 @@ func (c *Conn) ReadHeader() (*Header, error) {
 	if err := h.Body.UnmarshalBinary(body); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


kisom:
LGTM.

PR #120, opened 2016-09-06
remove gokeyless-tester and OpActivate

lziest:
we don't need a web service to test keyserver


lziest:
@kisom @elithrar @Bren2010 @jkroll-cf please review

kisom:
From a codebase perspective, the diff LGTM. As for removing activations and the keyless tester, I'll defer to @elithrar on that.

lziest:
The activation through certificate api is untouched. I am specifically wanting to delete OpActivate because that server would return a positive response if it is waiting for a certificate and it will return errors when it finishes activation. Also, when a server starts without certificate, OpActivate won't work because it is impossible to establish a TLS connection to a server yet.

OpActivate works like a ping, and so has no clear reason why it exists.

Bren2010:
LGTM

lziest:
@elithrar unless you protest, I will merge it at the end of the day.

elithrar:
@lziest No issues here.

PR #122, opened 2016-09-07
ignore key digest in the official keyserver and client implementation

lziest:
- we still keep digest in the protocol for backward compatibility
- we remove digest usage in server and client


lziest:
more deadlock plumbing

lziest:
the multi-thread server sometimes hangs.

kisom:
LGTM.

PR #126, opened 2016-09-09
conn.DoOperation (by its listen subroutine) should have timeout

lziest:
- the current communication model is as following:
  1. DoOperation writes a header, creates a channel and listen to the
     channel named by the Header ID.
  2. Each DoOperation will attempt a read, but the obtained Header by
     DoOperation call X may be for DoOperation call Y. And so X will
     still be waiting and hopefully another read attempt will return
     its result.
  3. DoOperation may forever hang on the channel. We need to timeout
     such hang.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -124,21 +124,7 @@ func (c *Conn) doRead() error {
 }
 
 // listenResponse attempts to read a response with the appropriate ID, blocking until it is available.
-func (c *Conn) listenResponse(id uint32) (*Header, error) {
-	c.mapMutex.Lock()
-	ch, ok := c.listeners[id]
-	if !ok {
-		ch = make(chan *Header, 1)
-		c.listeners[id] = ch
-	}
-	c.mapMutex.Unlock()
-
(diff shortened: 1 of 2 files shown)


lziest:
@kisom @Bren2010 @jkroll-cf please review. I am finally confident the channel usage is now correct.

conn.go
```diff
@@ -124,21 +124,7 @@ func (c *Conn) doRead() error {
...
-	ch, ok := c.listeners[id]
-	if !ok {
-		ch = make(chan *Header, 1)
-		c.listeners[id] = ch
-	}
-	c.mapMutex.Unlock()
-
-	defer func() {
-		c.mapMutex.Lock()
-		delete(c.listeners, id)
```
lziest: this is the culprit

kisom:
LGTM.

lziest:
need to a little bit more

lziest:
done, need to create the hashmap channel entry before the write operation is done

PR #127, opened 2016-09-12
remove unix socket file before starting to use it

lziest:


server/server.go
```diff
@@ -360,6 +361,7 @@ func (s *Server) ListenAndServe() error {
 // UnixListenAndServe listens on the Unix socket address and handles keyless requests.
 func (s *Server) UnixListenAndServe() error {
```
kisom: I think instead you should call `defer os.Remove(s.UnixAddr)` and have this occur as part of the cleanup.

Bren2010: You should never need to call os.Remove(socket). Calling Close() on a Unix socket is what removes it.

If possible, you should Close() the socket at cleanup instead of delete it at startup, but sometimes this is unavoidable. Optimal solution is to have circus/systemd manage sockets.

lziest: I found the test leaks the unix socket. But I haven't figured out a way to contain that.

kisom: Sorry, I meant to say defer the close.

@lziest How does it leak the socket?

lziest:
will create a new PR to address socket file leak.

PR #128, opened 2016-09-12
keyless client should retry on application level

lziest:
- in case remote server is down and Remote's PingAll has not yet executed,
  we need application-level retry on connection read/write error. Here it is,
  the remote-based private key would retry once if first attempt is bad


lziest:
@kisom can you take a look at this?

kisom:
LGTM.

PR #129, opened 2016-09-12
DO NOT MERGE - remove Client.remotes

FiloSottile:
The serverName -> *Remote mapping is a LRU, even if there won't be many
serverNames. Instead the SKI -> *Remote mapping is a simple map, from
which nothing is never removed. It will unavoidably grow in size the
more certs are observed.

One could obviously just use an LRU, but it got me thinking about why we
have that indirection anyway, when we could instead just stick the
serverName on the PrivateKey, and rely on the serverName -> *Remote LRU.

A lot of removed lines seem to suggest this simplifies things.

This is NOT a finished patch (haven't touched the tests), but it's
easier to talk over code.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -13,7 +13,6 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"sync"
 	"time"
 
 	"github.com/cloudflare/cfssl/helpers"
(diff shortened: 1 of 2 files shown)


lziest:
this is a good idea, it is what I am going to do next. The remote cache was once a map, I changed it, and prepare to change the server lookup dynamically next.

client/client.go
```diff
@@ -154,39 +136,30 @@ func (c *Client) DialDefault() (*Conn, error) {
...
-	cachedRemote := false
 	v, stale := c.remoteCache.Get(server)
 	if v != nil && !stale {
 		if r, ok := v.(Remote); ok {
-			c.remotes[ski] = r
-			cachedRemote = true
+			return r, nil
 		} else {
-			log.Error("failed to convert cached remote")
+			panic("failed to convert cached remote")
```
lziest: although it is unlikely, but I generally don't want to panic. We can simply fail gracefully.

lziest:
please proceed to smooth out API references.

client/client.go
FiloSottile: this should not be unlikely, but simply impossible unless there is a programmer error in adding (and guru tells me there's not), so it's the kind of place where panic() is appropriate IMHO, but I know that's subjective

FiloSottile:
(y)

lziest:
superceded by #143

PR #130, opened 2016-09-12
Don't merge - record listeners and properly close them when necessary

lziest:


lziest:
fixed unix socket file leak

server/server.go
```diff
@@ -117,6 +117,11 @@ type Server struct {
 	stats *statistics
 	// GetCertificate is used for loading certificates.
 	GetCertificate GetCertificate
+
+	// UnixListener is the listener serving unix://[UnixAddr]
```
kisom: Where is this used elsewhere in this file?

lziest: in UnixListenAndServe

lziest: It seems to me that we need to take care the unix listener because `defer l.Close()` never gets executed in goroutine when program exits normally. The tcp listener fares better as port will be freed but unix listener will hold the fd.

client/remote_test.go
```diff
@@ -140,7 +134,16 @@ func TestMain(m *testing.M) {
 		log.Fatal(err)
 	}
 
-	os.Exit(m.Run())
+	ret := m.Run()
+	// clean up server listeners
+	if s.UnixListener != nil {
```
kisom: Does the [`Serve`](https://github.com/cloudflare/gokeyless/pull/130/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bL336) function not clean these up properly?

server/server.go
```diff
@@ -336,11 +341,12 @@ func (s *Server) handleReq(conn *gokeyless.Conn, ch chan *gokeyless.Header) {
 func (s *Server) Serve(l net.Listener) error {
 	defer l.Close()
```
lziest: this never gets executed when program normally exits and one (or more) goroutine is blocked on l.Accept(). Will need to figure out why

lziest:
Don't merge, as I am figuring out the goroutine closure.

PR #131, opened 2016-09-13
client: Split off part of RegisterPublicKeyTemplate into RegisterSKI.

Bren2010:


lziest:
lgtm, I will merge now. @FiloSottile sorry if it introduces conflicts to your PR.

FiloSottile:
Eh, it's not just the conflict, but "RegisterSKI" as a name makes little sense after #131. But w/e, Register in general will make little sense as a prefix anyway.

PR #140, opened 2016-09-21
Allow the server to not respond to some requests.

Bren2010:
Allowing the gokeyless server to panic w/o dying means that some requests will not get responses, and should timeout.

Imagine you call "DoOperation" twice. The first request causes a panic and the second is successful. The first DoOperation reads the second response and puts it into the channel and blocks on waiting for its own response, eventually timing out. The second DoOperation will block on reading another header from the socket. If a third DoOperation happens and gets a successful response, the second DoOperation will unblock and return its response, and the third DoOperation will hang until a fourth DoOperation. And on...

### DIFF
--- a/client/remote.go
+++ b/client/remote.go
@@ -84,20 +84,17 @@ func healthchecker(conn *Conn) {
 
 	for {
 		time.Sleep(backoff.Duration())
-		if !conn.Conn.IsClosed() {
-			err := conn.Ping(nil)
-			if err != nil {
-				log.Debug("health check ping failed:", err)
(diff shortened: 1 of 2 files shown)


lziest:
To my understanding this is "Decouple read/write in DoOperation".

lziest:
lgtm

lziest:
Hmm, I don't quite understand the case described above. Suppose 1st DoOperation captures 2nd DoOperation response, 2nd DoOperation will unblock immediately?

PR #141, opened 2016-09-22
update ErrKeyNotFound messaging

lziest:


lziest:
@Bren2010 need to figure out the test error.

lziest:
@Bren2010 I think I fixed it, check out the last commit.

Bren2010:
Why does the last commit fix the tests? If you close the connection, the reader goroutine will stop, close again, and remove from connection pool.

PR #142, opened 2016-09-26
DefaultKeystore log SKI for each keyless operation

lziest:
- plus we adjusted some logging style when we are at it.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -55,7 +55,7 @@ func (keys *DefaultKeystore) Add(op *gokeyless.Operation, priv crypto.Signer) er
 
 	keys.skis[ski] = priv
 
-	log.Debugf("Adding key with SKI: %02x", ski)
+	log.Debugf("add key with SKI: %02x", ski)
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -68,11 +68,12 @@ func (keys *DefaultKeystore) Get(op *gokeyless.Operation) (crypto.Signer, bool)
 	if ski.Valid() {
 		priv, found := keys.skis[ski]
 		if found {
+			log.Infof("fetch key with SKI: %02x", ski)
 			return priv, found
 		}
 	}
 
-	log.Infof("Couldn't look up key for %s.", op)
+	log.Errorf("couldn't fetch key for %s.", op)
```
Bren2010: Whoa, won't this kill the server?

lziest: do you mean that will be a lot of errors? Or a single errorf will kill the server?

Bren2010: Sorry, nvm. I was used to Errorf being Printf followed by os.Exit(1). cfssl reserves Fatalf for that.

PR #143, opened 2016-09-27
Remove Client.remotes

lziest:
- Client.remotes serves as a mapping from SKIs to remotes, which has no real usage
- remove client Dial(SKI) as we can directly dial with remote
- rename RegisterKey* to NewRemoteSigner*
- much cleaner tests


lziest:
@FiloSottile please review.

Issue #144, opened 2016-09-27
Remote interface could have an additional method: PingAll()


lziest:
PingAll() basically dials all ip addresses for a given remote at one time to determine if there is at least one usable established connection. This would be useful for proxying requests through a extremely unreliable network to a upstream keyserver cluster.

PR #145, opened 2016-09-28
Make PingAll part of Remote interface, address #144

lziest:
- PingAll now uses channels and goroutines to collect ping measurements.
- rename item to mRemote


lziest:
@Bren2010 can you review?

lziest:
squashed and rebased.

lziest:
maybe I can use the size of channel buffer to limit the concurrency for background health check.

lziest:
@Bren2010  now the concurrency of PingAll can be controlled. For healthcheck, concurrency is 1, for the purpose of finding one usable keyserver, I can set concurrency to a high number.

PR #146, opened 2016-09-29
fix the read race on g.remotes

lziest:
see https://travis-ci.org/cloudflare/gokeyless/jobs/163552037
it's very benign race though, since len(g.remotes) seldom changes.


lziest:
given the simplicity of this change and that tests are passing, I will merge this right away.

PR #150, opened 2016-11-01
simplify NewServer

lziest:
- Server's exported 'Addr' and 'UnixAddr' fields can be safely decoupled
  from Server instance initialization


lziest:
actually this feels very heavy handed.

PR #153, opened 2016-11-08
Cleanup Go Report Card warnings

gliptak:
Correcting a number of warning from https://goreportcard.com/report/github.com/cloudflare/gokeyless


conn.go
```diff
@@ -9,6 +9,7 @@ import (
 	"time"
 )
 
+// Default timeout (10 seconds)
```
kisom: I think this should be changed to

```
// OperationTimeout is the default timeout value.
```

server/server.go
```diff
@@ -390,6 +390,7 @@ func (s *Server) UnixListenAndServe() error {
 	return errors.New("can't listen on empty path")
 }
 
+// Close listeners
```
kisom: Comments should be complete sentences (golint might complain about this). I think maybe this should be

```
Close shuts down the listeners.
```

client/keys.go
```diff
@@ -14,6 +14,25 @@ import (
 	"github.com/cloudflare/gokeyless"
 )
 
+func mapRSACrypto(hash crypto.Hash) gokeyless.Op {
```
kisom: This should probably be a map, instead of a function.

client/remote_test.go
```diff
@@ -203,7 +203,7 @@ func TestSlowServer(t *testing.T) {
 
 	// After ping checks, 1st remote must be the normal server.
 	firstRemote := g.remotes[0]
-	conn, err := firstRemote.Dial(c)
```
kisom: Rather than assigning to `_`, this should check the error and fail the test if it's not `nil`.

server/server.go
gliptak: done

client/remote_test.go
gliptak: done

client/keys.go
gliptak: done

gliptak:
@kisom Thank you for the detailed review. I mad the changes suggested

client/keys.go
```diff
@@ -14,6 +14,24 @@ import (
 	"github.com/cloudflare/gokeyless"
 )
 
+const rsaCrypto = map[crypto.Hash] gokeyless.Op {
```
lziest: can't `const`, should be `var`, and please group `const` or `var`

gliptak: What do you mean by "group"?

lziest: I mean like this https://golang.org/doc/effective_go.html#variables

gliptak: done

lziest:
lgtm

kisom:
Thanks, @gliptak.

PR #155, opened 2016-12-02
remove latency stats in tests result

lziest:
it's way too visual confusing to have mostly useless latency stats, while tyring to focus on success/failure


lziest:
fix typo in commit message

PR #160, opened 2017-02-22
Add OpSeal/OpSealer and Server.Sealer

FiloSottile:
These will be used GetCertificate-style to do session tickets from out-of-process.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -119,8 +119,7 @@ type Server struct {
 	// GetCertificate is used for loading certificates.
 	GetCertificate GetCertificate
 	// Sealer is called for Seal and Unseal operations.
-	// Sealer can return a gokeyless.Error to send a custom error code.
-	Sealer func(*gokeyless.Operation) ([]byte, error)
+	Sealer Sealer
 
 	// UnixListener is the listener serving unix://[UnixAddr]
 	UnixListener net.Listener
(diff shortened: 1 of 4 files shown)


lziest:
won't review it very soon for today or tomorrow.

grittygrease:
This generally looks good and doesn't change existing code paths much.

server/server.go
```diff
@@ -118,6 +118,9 @@ type Server struct {
 	stats *statistics
 	// GetCertificate is used for loading certificates.
 	GetCertificate GetCertificate
+	// Sealer is called for Seal and Unseal operations.
+	// Sealer can return a gokeyless.Error to send a custom error code.
+	Sealer func(*gokeyless.Operation) ([]byte, error)
```
grittygrease: This function signature could be made into a type as GetCertificate below.

lziest:
@FiloSottile apologies for low availability. If you want, you can merge it yourself given CI is green.

server/server.go
lziest: Maybe make it a interface would make more sense. As it force implementor to have both Seal and Unseal operation.

FiloSottile: I considered a named type, but there is no meaningful advantage AFAICT, and it makes the definition stutter.

As for the interface, I see the idea, but OTOH having two identical methods is weird.

If you don't have strong opinions, I'd go with what we have, otherwise I'll change it to an interface.

lziest: hmm, unless we are doing xor. Seal and Unseal being identical?

FiloSottile: ah, yeah, didn't mean they behave identically, just that they have the same method signature. Essentially I don't see it as much different to switch inside the function.

But again, weak preference.

lziest: I understand, my preference is that we have a `Sealer` interface, has `Seal` and `Unseal` functions, and in the switch block below, we actually match Seal with Opseal, Unseal with Opunseal. My feeling is that that is easier to make changes later on, and less error prone

FiloSottile: No problem, I'll make the change.

lziest:
LGTM, the original `Body *Operation` definition makes it hard at times to understand. Maybe a new naming PR is coming :-)

PR #161, opened 2017-02-28
reduce tcp listener's connection timeout to restrict goroutine growth

lziest:


lziest:
@grittygrease @FiloSottile @Bren2010 I am thinking about cutting over using long-lived Unix connection locally and so we will not keep thousands of TCP connection open for one hour for only a handful incoming requests per connection, which could be the use case of customer keyserver.

PR #163, opened 2017-03-01
Zi/fix nil func

lziest:


lziest:
@Bren2010 I touched your code a little bit, can you review?

Issue #162, opened 2017-03-01
The Header definition is misleading


lziest:
We should consider rename the Header structure, because it is actually define the entire operation packet. Header has everything now.

PR #164, opened 2017-03-01
rename Header to Packet

lziest:


lziest:
squashed

lziest:
address #162

PR #165, opened 2017-06-08
Server-specific worker pools

Bren2010:
This is just an idea for how to have pre-prepared workers pools and long lived connections. I haven't really vetted the code.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -23,11 +23,6 @@ import (
 
 var keyExt = regexp.MustCompile(`.+\.key`)
 
-const (
-	UnixConnTimeout = time.Hour
-	TCPConnTimeout  = time.Second * 30
-)
-
 // Keystore is an abstract container for a server's private keys, allowing
 // lookup of keys based on incoming `Operation` requests.
 type Keystore interface {
@@ -180,7 +175,7 @@ func NewServerFromFile(certFile, keyFile, caFile, addr, unixAddr string) (*Serve
 	return NewServer(cert, keylessCA, addr, unixAddr), nil
 }
 
-func (s *Server) handle(conn *gokeyless.Conn, timeout time.Duration) {
+func (s *Server) handle(conn *gokeyless.Conn) {
 	defer conn.Close()
 	log.Debug("Handling new connection...")
 
@@ -195,7 +190,7 @@ func (s *Server) handle(conn *gokeyless.Conn, timeout time.Duration) {
 	var connError error
 	var h *gokeyless.Packet
 	for connError == nil {
-		conn.SetDeadline(time.Now().Add(timeout))
+		conn.SetDeadline(time.Now().Add(time.Hour))
 
 		if h, connError = conn.ReadPacket(); connError != nil {
 			break
@@ -209,7 +204,6 @@ func (s *Server) handle(conn *gokeyless.Conn, timeout time.Duration) {
 	} else if err, ok := connError.(net.Error); ok && err.Timeout() {
 		log.Debugf("server closes connection due to timeout: %v\n", err)
 	} else {
-		s.stats.logConnFailure()
 		log.Errorf("connection error: %v\n", connError)
 	}
 }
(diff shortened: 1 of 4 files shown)


lziest:
:) feels like writing some part of nginx.

Bren2010:
Yeah, it might be better to remove the worker pools altogether and go back to what you were doing before. =/ What do you think?

lziest:
I haven't got time to reason about the channel growth, I will review soon

server/server.go
```diff
@@ -180,55 +183,54 @@ func NewServerFromFile(certFile, keyFile, caFile, addr, unixAddr string) (*Serve
...
+		ch <- req{conn, h}
+		s.workerMu.Unlock()
 	}
 
 	if connError == io.EOF {
 		log.Debug("connection closed by client")
 	} else if err, ok := connError.(net.Error); ok && err.Timeout() {
 		log.Debugf("server closes connection due to timeout: %v\n", err)
 	} else {
-		s.stats.logConnFailure()
```
lziest: curious about this deleted line.

server/server.go
```diff
@@ -392,15 +394,27 @@ func (s *Server) handleReq(conn *gokeyless.Conn, ch chan *gokeyless.Packet) {
...
+	}()
+
 	for {
 		c, err := l.Accept()
 		if err != nil {
 			log.Error(err)
 			return err
 		}
-		go s.handle(gokeyless.NewConn(tls.Server(c, s.Config)), timeout)
+		go s.handle(gokeyless.NewConn(tls.Server(c, s.Config)), ch)
```
lziest: so I believe this was causing goroutine to grow out of control, since each `go s.handle()` will spawn a goroutine which is blocked for one hour for each incoming connection. And when a lot of test machines trying to do just one key signing request, the server just has a lot of goroutine waiting for nothing.  The original fix, arguably ugly, is to has a smaller timeout for tcp connection, so "we hope" some goroutine will exit naturally after timeout. Frankly, I do feel that won't help much.

So in order to use the channel model so we don't create unlimited number of goroutine, I think we need to have this master thread block waiting for a available worker,  so, simply stuffing new (conn, pakcket) into a channel will do that, and we can eliminate this `go s.handle call()`. What do you think?

lziest: TLDR, do 's.handle()' instead of 'go s.handle()', then I think we are gold.

lziest: TLDR, I think we need to rewrite a bigger part of it. The dead-loop polling packet from the connection probably is the root cause we can't scale.

lziest: The more i think about it, the more I feel like we are implementing select(2)

Bren2010: The goroutine will stop blocking and exit if the connection closes. Plus, for each connection, we only spawn a new goroutine, and goroutines are much cheaper than threads.

I did an experiment and made 5k simultaneous connections to a keyserver -- without this PR the server crashes, and with this PR memory stabilizes at 70mb.

server/server.go
Bren2010: Good catch, will fix.

server/server.go
lziest: I guess it might be due to the 10x factor in the `handle()`? Can you try 50k to confirm?

server/server.go
```diff
@@ -180,28 +183,27 @@ func NewServerFromFile(certFile, keyFile, caFile, addr, unixAddr string) (*Serve
...
 }
 
-func (s *Server) handle(conn *gokeyless.Conn, timeout time.Duration) {
+func (s *Server) handle(conn *gokeyless.Conn, ch chan req) {
 	defer conn.Close()
 	log.Debug("Handling new connection...")
 
-	ch := make(chan *gokeyless.Packet, 10)
-	defer close(ch)
-	for i := 0; i < 10; i++ {
```
lziest: that's the 10x factor I am talking about.

server/server.go
Bren2010: I tried with 25k and got 300-290mb memory. 50k connections breaks for reasons that I don't think are keyless's fault (being too close to 2^16).

lziest:
@Bren2010 can I fix benchmarking keyless test first then I can merge this with confidence?

Bren2010:
Sounds good to me.

lziest:
we would merge this, seems like we would do benchmarking very soon.

PR #166, opened 2017-06-13
Use tls.Config.Clone, rather than copyTLSConfig.

Bren2010:
Something cool I noticed :)


Bren2010:
Looks like it's new to go1.8. :(

lziest:
we can move to 1.8 very soon.

PR #168, opened 2017-06-15
Add micro- and macro-benchmarks for throughput and latency

joshlf:
This PR adds some microbenchmarks for cryptographic computations and a macrobenchmarking tool to measure latency and bandwidth characteristics of a Keyless server.

### DIFF
--- a/cmd/bench/internal/client/client.go
+++ b/cmd/bench/internal/client/client.go
@@ -1,0 +1,246 @@
+package client
+
+import (
+	"fmt"
+	"math"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// A LatencyBucket represents one bucket in a latency histogram, recording the
+// number of requests which took in the range [From, Until].
+type LatencyBucket struct {
+	From, Until time.Duration
+	Count       uint64
+}
+
+// PrintHistogram prints a histogram of the given buckets to stdout. The buckets
+// are assumed to be non-overlapping intervals sorted in ascending order.
+func PrintHistogram(buckets []LatencyBucket) {
+	// To save screen space - especially because RunLatencyClients always returns
+	// a histogram whose largest bucket goes to the maximum time.Duration value -
+	// represent time.Duration(math.MaxInt64) as "-" instead of formatting the
+	// actual time.
+	durString := func(d time.Duration) string {
+		if d == time.Duration(math.MaxInt64) {
+			return "-"
+		}
+		return fmt.Sprint(d)
+	}
+
+	max := uint64(0)
+	firstNonzero := -1 // first bucket with Count > 0
+	lastNonzero := 0   // last bucket with Count > 0
+	maxFromLen := 0    // maximum length of string-formatted From field
+	maxUntilLen := 0   // maximum length of string-formatted Until field
+	for i, b := range buckets {
+		if b.Count > max {
+			max = b.Count
+		}
+		if b.Count > 0 {
+			if firstNonzero == -1 {
+				firstNonzero = i
+			}
+			lastNonzero = i
+
+			// Only count the length of the durations we're going to print - thus, do
+			// this logic inside this if block. The input could have 0-count buckets
+			// in between other non-zero-count buckets, but as long as the buckets are
+			// ascending in their from/until durations, the later buckets will always
+			// have equal- or greater-length duration strings than the current one, so
+			// skipping intermediate zero-count buckets is acceptable.
+			fromlen, untillen := len(durString(b.From)), len(durString(b.Until))
+			if fromlen > maxFromLen {
+				maxFromLen = fromlen
+			}
+			if untillen > maxUntilLen {
+				maxUntilLen = untillen
+			}
+		}
+	}
+
+	maxCountLen := len(fmt.Sprint(max))
+
+	fmtstr := fmt.Sprintf("[%%%vv, %%%vv) %%%vv | %%v\n", maxFromLen, maxUntilLen, maxCountLen)
+	fmax := float64(max)
+	for _, b := range buckets[firstNonzero : lastNonzero+1] {
+		perc := float64(b.Count) / fmax
+		height := int(perc * 70)
+		fmt.Printf(fmtstr, durString(b.From), durString(b.Until), b.Count, strings.Repeat("=", height))
+	}
+}
+
+// A BandwidthClient is a client used for bandwidth measurements. It must be
+// safe for Dispatch and Complete to be called concurrently from different
+// goroutines, although neither method needs to be safe for other concurrent
+// calls to itself. In other words, at most one call to each of the two methods
+// will be in progress at any given time.
+type BandwidthClient interface {
+	// Dispatch dispatches a new request, but does not wait for it to complete.
+	Dispatch()
+
+	// Complete waits for the next request to complete. It must wait for whatever
+	// request is the first to complete rather than waiting for a particular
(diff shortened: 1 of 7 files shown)


cmd/bench/main.go
```diff
@@ -0,0 +1,225 @@
...
+	return &gokeyless.Operation{
+		Opcode:  params.Opcode,
+		Payload: randBytes(params.PayloadSize),
+		SKI:     SKI,
+	}
+}
+
+func randBytes(n int) []byte {
+	b := make([]byte, n)
+	_, err := io.ReadFull(rand.New(rand.NewSource(time.Now().UnixNano())), b)
```
lziest: the reason behind new seeded rand call for each randBytes()?

joshlf: Nope, no good reason, although I generally prefer to do it this way in case I end up using `randBytes` for something where a different seed each time is important (admittedly, that's unlikely to happen in this code).

server/microbench_test.go
```diff
@@ -0,0 +1,285 @@
...
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	"io"
+	"runtime"
+	"sync"
+	"testing"
+
+	"github.com/cloudflare/gokeyless/internal/test/params"
+	"github.com/joshlf/testutil"
```
Bren2010: Does this need to be vendored?

cmd/bench/main.go
lziest: I have concerns though about that repeatedly reseeding a new rand stream is no better than using one rand stream. For efficiency, we can generate the source once each run and use it for the run time. What do think?

cmd/bench/internal/client/client.go
```diff
@@ -0,0 +1,246 @@
...
+	latencyCounts := make([]uint64, nbuckets+2)
+	for !stop() {
+		dur := c.Do()
+		if dur < bottom {
+			latencyCounts[0]++
+		} else if dur > top {
+			latencyCounts[nbuckets-1]++
+		} else {
+			bucketIdx := int((dur - bottom) / skip)
+			latencyCounts[bucketIdx]++
```
Bren2010: bucketIdx+1, right?

cmd/bench/internal/client/client.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+	}
+
+	buckets := make([]LatencyBucket, nbuckets+2)
+	buckets[0].Until = bottom - 1
```
Bren2010: I'm not sure about the -1ns. Since the histogram claims half-open interval: [x, y). Line 242 as well.

cmd/bench/main.go
joshlf: This only gets called once at the beginning of the program for initialization, so I wouldn't worry about performance.

server/microbench_test.go
joshlf: @lziest has concerns about vendoring too much, and this is only in a test, so I figured it's fine to just let people use 'go get' to run the tests. There are a number of other test dependencies that are also not vendored. I'd also be happy to vendor it, though.

cmd/bench/internal/client/client.go
joshlf: Good call.

cmd/bench/internal/client/client.go
joshlf: The buckets are actually fully closed (see the comments on `LatencyBucket`) so that you can represent the maximum duration. It's kind of a stupid reason tbh, and I'd be happy to just make them half-open like you suggested and document that we don't support `time.Duration(math.MaxInt64)`.

Bren2010: Then is the format on line 66 inconsistent?  I think it would be a good idea to have them be half-open.

joshlf: You're right, that formatting is inconsistent. I'll just make them half-open.

cmd/bench/main.go
lziest: but for each 'randByte' call, a new seeded source is created, right?

joshlf: Yes, but `randBytes` is only called in `makeXXXSignOp`, which is only called at initialization time.

cmd/bench/main.go
```diff
@@ -0,0 +1,225 @@
...
+	config := cli.Config.Clone()
+	config.ServerName = server
+	conn, err := tls.Dial("tcp", server+":"+port, config)
+	if err != nil {
+		panic(err)
+	}
+	return conn
+}
+
+func makeRSASignOp(params params.RSASignParams, SKI gokeyless.SKI) *gokeyless.Operation {
```
lziest: this is never used?

cmd/bench/main.go
lziest: makeECDSASignOp is called in the loop. What do you feel about using 'crypto/rand'? I just want to keep the best practice.

joshlf: That loop only runs once per worker goroutine, so it's not a big deal. But yeah, crypto/rand would be fine.

cmd/bench/main.go
joshlf: I may extend the tool to let the user specify what kind of signing operation they want to test with, in which case I'll use that.

joshlf: Update: I just did :)

cmd/bench/main.go
joshlf: OK, fixed.

cmd/bench/main.go
```diff
@@ -0,0 +1,252 @@
...
+		Payload: randBytes(params.PayloadSize),
+		SKI:     SKI,
+	}
+}
+
+func makeECDSASignOp(params params.ECDSASignParams, SKI gokeyless.SKI) *gokeyless.Operation {
+	return &gokeyless.Operation{
+		Opcode:  params.Opcode,
+		Payload: randBytes(params.PayloadSize),
+		SKI:     SKI,
```
Bren2010: It would be good to be able to set the SNI and ServerIP fields.

joshlf: What happens if both SKI and SNI are set? Seeing as this is just a benchmarking tool, as long as SKI is sufficient, I'm happy only supporting SKI, although I suppose SNI might make it easier for a human to specify which cert they want.

cmd/bench/internal/client/client.go
```diff
@@ -0,0 +1,250 @@
...
+	Do() time.Duration
+}
+
+// FuncLatencyClient is a function that implements the LatencyClient interface.
+type FuncLatencyClient func() time.Duration
+
+// Do invokes f.
+func (f FuncLatencyClient) Do() time.Duration { return f() }
+
+var _ LatencyClient = FuncLatencyClient(nil)
```
lziest: what's this for?

Bren2010: So the code doesn't compile if FuncLatencyClient doesn't implement the LatencyClient interface.

server/microbench_test.go
lziest: you can vendor it at cmd/vendor

joshlf: Done.

lziest:
feel free to merge it

joshlf:
@lziest if you add me to the repo, I will.

lziest:
@joshlf probably need to add yourself to Security team on Github. Maybe @grittygrease would know how to do file a ticket.

PR #170, opened 2017-06-16
Refactor package layout

joshlf:
Made the following changes:
* Move top-level `gokeyless` package to `internal/protocol`
* Move the connection handling functionality that used to be in `gokeyless` into its own `conn` package
* Move `server` package to `internal/server`
* Clean up serialization and deserialization code (now in `internal/protocol`)


lziest:
client/remote_test.go:16:2: cannot find package "github.com/cloudflare/gokeyless/server"

lziest:
I think moving server into internal is not desirable, we still vendor server package somewhere, unless you keep a public server/server.go which has only interface definition.

lziest:
Once tests passed, we can merge, but we need to make Keystore interface public and server.Server struct public later.

PR #173, opened 2017-07-06
Revert "Revert "reduce tcp listener's connection timeout to restrict …

Bren2010:
…goroutine growth""

This reverts commit 7fb7200eccdd5988f35c2bd4a6c1be4c5af9a299.

The go=1.6 tests fail because it looks like a dependency was rewritten to use contexts. Is that a blocker?


lziest:
should not be a blocker, will update travis.yaml and get this rebased.

PR #175, opened 2017-07-17
Rewrite server algorithm to improve performance, refactor APIs

joshlf:
## API modifications
* Move `server` and `protocol` packages out of `internal`
* Reduce API surface of `server` to be as minimal as possible, clean up API to be more usable

## Performance
* Redesigned server to use worker pool model
  * A fixed number of worker goroutines perform requested operations
  * Two separate pools - one for ECDSA ops, one for everything else
  * Each connection has a pair of reader/writer goroutines that submit jobs to the pool and read responses from the pool
* Pre-compute ECDSA random values


joshlf:
Note: There's a minor bug that I've been unable to track down yet where logging is always set to DEBUG regardless of what's set on the command line. If anyone knows what that might be, let me know.

tests/proxy_test.go
```diff
@@ -82,7 +82,7 @@ func TestTLSProxy(t *testing.T) {
 	time.Sleep(100 * time.Millisecond)
 
 	keys := server.NewDefaultKeystore()
-	s.Keys = keys
+	s.SetKeystore(keys)
```
lziest: In general we don't need setter func if the field is exported.

joshlf: The field isn't exported anymore. I made all of the fields in `Server` private.

lziest: apologies, the diff is hidden by default so I skipped that by scrolling

joshlf: np

lziest:
the log behavior I may know, I will come to you tomorrow.

internal/server/internal/ecdsa/ecdsa.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+// NOTE: This buffer is designed to ensure that, regardless of how code outside
+// of this package uses these types and methods, no random value will ever be
+// used more than once in an ECDSA signature. When modifying this code, be VERY
+// CAREFUL that this behavior is maintained.
+
+// A RandProvider is capable of producing random values for use in ECDSA
+// signatures. It cannot be implemented by types outisde of this package.
+type RandProvider interface {
```
lziest: if so, can we make the interface non-exported?

joshlf: We want this to be exported - it needs to be used outside the `ecdsa` package. But the `ecdsa` package is inside `internal`, so it's only accessible by other packages in `github.com/cloudflare/gokeyless`.

internal/server/internal/worker/worker.go
```diff
@@ -0,0 +1,136 @@
...
+func (p *Pool) Destroy() {
+	for i := 0; i < p.workers; i++ {
+		// Send the sentinal value (with the 'job' field as nil) indicating to quit;
+		// send one for each worker goroutine.
+		p.jobs <- Job{}
+	}
+	p.wg.Wait()
+}
+
+func (p *Pool) worker(w Worker) {
```
lziest: I prefer it to be inlined.

internal/server/internal/worker/worker.go
```diff
@@ -0,0 +1,136 @@
...
+// goroutines have quit. If p has already been destroyed, the behavior of
+// Destroy is undefined.
+func (p *IdlePool) Destroy() {
+	// Instruct the idle goroutines to quit. They will observe this change on the
+	// next loop iteration.
+	atomic.StoreUint32(&p.quit, 1)
+	p.wg.Wait()
+}
+
+func (p *IdlePool) worker(i Idler) {
```
lziest: I plan it to be inlined.

internal/server/internal/ecdsa/ecdsa.go
lziest: so far I haven't seen any use outside of ecdsa package.

server/internal/worker/worker.go
```diff
@@ -0,0 +1,136 @@
...
+// method, and once the Do method returns, its return value will be passed as
+// the argument to commit.
+func NewJob(job interface{}, commit func(result interface{})) Job {
+	if job == nil {
+		panic("nil job")
+	}
+	return Job{job: job, commit: commit}
+}
+
+// An Idler performs a unit of background work when Idle is called.
```
lziest: I need to understand this. Is it for perf testing?

joshlf: No, the Idler is used to compute ECDSA random values in the background.

internal/server/internal/worker/worker.go
joshlf: I tend to prefer it to be its own method so that it's very obvious what's happening when we call it on line 57. It also makes it easier to separate out the logic of synchronizing the workers (e.g., the `p.wg.Done()` call).

server/internal/worker/worker.go
lziest: so there is more change coming?

internal/server/internal/worker/worker.go
lziest: Pool.worker() is entirely different than Pool.worker() yet they share the same name, and I have no idea what is happening.

internal/server/internal/worker/worker.go
```diff
@@ -0,0 +1,136 @@
...
+// its own goroutine. The workers wait jobs to be submitted and execute those
+// jobs as they come in.
+func NewPool(workers ...Worker) *Pool {
+	p := &Pool{
+		jobs:    make(chan Job, 1024*1024),
+		workers: len(workers),
+	}
+
+	p.wg.Add(len(workers))
+	for i := 0; i < len(workers); i++ {
```
lziest: do you mind using idiomatic "for _, w := range workers"?

server/internal/worker/worker.go
joshlf: Yes, but nothing related to this. We create an idler to generate ECDSA random values [here](https://github.com/cloudflare/gokeyless/pull/175/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR455).

internal/server/internal/worker/worker.go
joshlf: lol I'm not sure why I didn't do that. I'll change it.

internal/server/internal/worker/worker.go
joshlf: What do you mean "Pool.worker() is entirely different than Pool.worker()"?

internal/server/internal/ecdsa/ecdsa.go
joshlf: Search for `buf_ecdsa` in `server.go` - I had to rename the import so it wouldn't conflict with Go's stdlib `ecdsa` package.

server/internal/worker/worker.go
lziest: In other words, idler is not idle. It's more like generic CPU worker. I personally dislike the naming.

server/internal/worker/worker.go
```diff
@@ -0,0 +1,136 @@
...
+type Idler interface {
+	// Idle performs a unit of background work.
```
lziest: I think it is a little bit misleading here to mention "a unit" and "background", "background" is done by goroutine, and "a unit" is not well defined.

server/internal/worker/worker.go
```diff
@@ -0,0 +1,136 @@
...
+type Pool struct {
+	jobs    chan Job
+	workers int // number of workers (not including idlers)
+	wg      sync.WaitGroup
+}
+
+// NewPool constructs a new Pool from the given workers. Each worker is run in
+// its own goroutine. The workers wait jobs to be submitted and execute those
+// jobs as they come in.
+func NewPool(workers ...Worker) *Pool {
```
lziest: I feel you just need `workers []Worker`

server/server.go
```diff
@@ -0,0 +1,863 @@
...
+				protocol.OpECDSASignSHA384, protocol.OpECDSASignSHA512:
+				ecdsaPool.SubmitJob(job)
+			default:
+				otherPool.SubmitJob(job)
+			}
+		}
+	}()
+
+	// writer
+	go func() {
```
lziest: I want this function not to be inlined.

server/server.go
```diff
@@ -0,0 +1,863 @@
...
+	// If the writer is currently blocked in a call to write, that call will
+	// return when the connection is closed. If the writer is currently blocked
+	// reading from the responses channel, that call will return when the nil
+	// value is sent. In either case, the writer will know to quit.
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	// reader
+	go func() {
```
lziest: I want this function not to be inlined.

server/internal/worker/worker.go
joshlf: Can you think of a better description? The idea behind calling it an `Idler` is that it does work on its own without any input (as opposed to the `Worker`s, which accept job definitions that are submitted from elsewhere).

server/internal/worker/worker.go
joshlf: Using varargs is more idiomatic, and it allows users to call it like `NewPool(a, b, c)` if they want.

server/internal/worker/worker.go
lziest: I would call it `type Runner interface { Run();}` so we say 'Runner' simply represents any type of CPU-bound work.

server/internal/worker/worker.go
lziest: agreed

server/internal/worker/worker.go
joshlf: OK fair enough. I'll rename it.

Bren2010:
- Two separate pools - one for ECDSA ops, one for everything else
- Pre-compute ECDSA random values

I don't see any discussion about empirical performance improvements?

joshlf:
@Bren2010 

> * Two separate pools - one for ECDSA ops, one for everything else
> * Pre-compute ECDSA random values
> 
> I don't see any discussion about empirical performance improvements?

Here's a histogram of latency with the current master taken using the `bench` tool:

```text
[ 240µs,  260µs)   1 | 
[ 260µs,  280µs)   6 | =
[ 280µs,  300µs)  26 | ====
[ 300µs,  320µs)  65 | ===========
[ 320µs,  340µs) 156 | ===========================
[ 340µs,  360µs) 326 | =========================================================
[ 360µs,  380µs) 400 | ======================================================================
[ 380µs,  400µs) 337 | ==========================================================
[ 400µs,  420µs) 229 | ========================================
[ 420µs,  440µs) 136 | =======================
[ 440µs,  460µs)  95 | ================
[ 460µs,  480µs)  78 | =============
[ 480µs,  500µs)  85 | ==============
[ 500µs,  520µs)  87 | ===============
[ 520µs,  540µs)  94 | ================
[ 540µs,  560µs) 113 | ===================
[ 560µs,  580µs) 132 | =======================
[ 580µs,  600µs) 148 | =========================
[ 600µs,  620µs) 132 | =======================
[ 620µs,  640µs) 102 | =================
[ 640µs,  660µs)  82 | ==============
[ 660µs,  680µs)  50 | ========
[ 680µs,  700µs)  56 | =========
[ 700µs,  720µs)  47 | ========
[ 720µs,  740µs)  53 | =========
[ 740µs,  760µs)  44 | =======
[ 760µs,  780µs)  38 | ======
[ 780µs,  800µs)  29 | =====
[ 800µs,  820µs)  20 | ===
[ 820µs,  840µs)  18 | ===
[ 840µs,  860µs)  13 | ==
[ 860µs,  880µs)   9 | =
[ 880µs,  900µs)  11 | =
[ 900µs,  920µs)   8 | =
[ 920µs,  940µs)   5 | 
[ 940µs,  960µs)   8 | =
[   1ms,      -)  84 | ==============
```

Here's a histogram taken with the code in this PR:

```text
[ 120µs,  140µs)   4 | 
[ 140µs,  160µs)  10 | =
[ 160µs,  180µs)  36 | ====
[ 180µs,  200µs) 175 | =======================
[ 200µs,  220µs) 323 | ===========================================
[ 220µs,  240µs) 395 | =====================================================
[ 240µs,  260µs) 515 | ======================================================================
[ 260µs,  280µs) 485 | =================================================================
[ 280µs,  300µs) 363 | =================================================
[ 300µs,  320µs) 238 | ================================
[ 320µs,  340µs) 163 | ======================
[ 340µs,  360µs) 129 | =================
[ 360µs,  380µs) 100 | =============
[ 380µs,  400µs)  77 | ==========
[ 400µs,  420µs)  37 | =====
[ 420µs,  440µs)  30 | ====
[ 440µs,  460µs)  38 | =====
[ 460µs,  480µs)  28 | ===
[ 480µs,  500µs)  21 | ==
[ 500µs,  520µs)  22 | ==
[ 520µs,  540µs)  12 | =
[ 540µs,  560µs)  15 | ==
[ 560µs,  580µs)  13 | =
[ 580µs,  600µs)  12 | =
[ 600µs,  620µs)  10 | =
[ 620µs,  640µs)   6 | 
[ 640µs,  660µs)   6 | 
[ 660µs,  680µs)   7 | 
[ 680µs,  700µs)   6 | 
[ 700µs,  720µs)   6 | 
[ 720µs,  740µs)   5 | 
[ 740µs,  760µs)  10 | =
[ 760µs,  780µs)   6 | 
[ 780µs,  800µs)   3 | 
[ 800µs,  820µs)   4 | 
[ 820µs,  840µs)   2 | 
[ 840µs,  860µs)   6 | 
[ 860µs,  880µs)   1 | 
[ 880µs,  900µs)   3 | 
[ 900µs,  920µs)   4 | 
[ 920µs,  940µs)   4 | 
[ 940µs,  960µs)   2 | 
[   1ms,      -)  88 | ===========
```

Note that there are more improvements planned on top of this, but I wanted to make sure at least this was merged first so we're all working on the same code and not completely diverged.

Bren2010:
What are the future improvements? This PR is too complicated to save 200 microseconds.

joshlf:
@Bren2010 

> What are the future improvements? This PR is too complicated to save 200 microseconds.

Swapping out for a faster channel implementation.

More importantly, though, the code is just better organized and cleaned up. For example, there are tons of concurrency edge cases that are easy to get bitten by, and the current version isn't careful about that at all (including concurrent unsychronized reads/writes to shared variables). I smoked out a bunch of subtle concurrency edge cases here. To get a sense of what I'm talking about, take a look at the comments in internal/worker/client.go, for example. I'd be happy to discuss, but the TLDR is that this went through multiple revisions because bugs kept cropping up. I think it's important to make sure we get the concurrency story right because otherwise we're asking to get bitten in production.

joshlf:
Note: There's still one outstanding race condition, but it's just in a piece of code that's used only in a test (server/internal/worker/client_test.go). I'll fix it tomorrow.

server/internal/worker/worker.go
joshlf: How about having the pool be named `BackgroundPool` because it executes work in the background and doesn't accept requests from elsewhere in the system?

server/server.go
```diff
@@ -0,0 +1,825 @@
...
+		log.Errorf("Worker %v: %s: Signing error: %v\n", w.name, protocol.ErrCrypto, err)
+		w.s.stats.logInvalid(requestBegin)
+		return makeErrResponse(pkt.ID, protocol.ErrCrypto)
+	}
+
+	w.s.stats.logRequestDuration(requestBegin)
+	return makeRespondResponse(pkt.ID, sig)
+}
+
+type backgroundWorker struct {
```
lziest: I think`entropyWorker` is more appropriate. Every worker is background worker at some time.

server/server.go
```diff
@@ -0,0 +1,825 @@
...
+// connection from clients over the network. See the documentation in the worker
+// package for details.
+type client struct {
+	conn net.Conn
+	// name used to identify this client in logs
+	name                 string
+	timeout              time.Duration
+	ecdsaPool, otherPool *worker.Pool
+	// used by the Log method
+	err sync.Once
```
lziest: the `err` name is confusing to me

server/server.go
```diff
@@ -0,0 +1,825 @@
...
+func (c *client) Log(err error) {
+	// Use a sync.Once so that only the first goroutine to encounter an error gets
+	// to log it. This avoids the circumstance where a goroutine encounters an
+	// error, logs it, and then closes the network connection, which causes the
+	// other goroutine to also encounter an error (due to the closed connection)
+	// and spuriously log it.
+	//
+	// We also use this to allow Destroy to block the reader or writer from
+	// logging anything at all by calling Log(nil).
+	c.err.Do(func() {
```
lziest: wait...does this mean the client only log once?

joshlf: Yes - see the accompanying comment.

server/server.go
joshlf: Would `logErr` be better?

server/server.go
joshlf: OK

joshlf: How about `randGenWorker`?

server/server.go
lziest: so if a client is having multiple requests and one error will stop it from logging? I can't understand this.

lziest: I feel this is over-engineered.

joshlf: These are all connection errors, not logic errors (e.g., sending back `OpError` in response to a request isn't logged here), so a single error means that the connection should be torn down.

joshlf: I'll modify the comment to be clearer about that.

server/internal/worker/client.go
```diff
@@ -0,0 +1,383 @@
...
+// A Client is a server-side handle on a connection to a client. It is capable
+// of reading job requests from the client, writing responses to the client, and
+// shutting down the connection.
+//
+// Since Clients can be backed by TCP connections, it is acceptable for GetJob
+// to return false (indicating that there are no jobs to be done - in the case
+// of a TCP connection, caused by a half-close of the connection) while
+// SubmitResult continues to return true (because writing to the connection is
+// still allowed).
+type Client interface {
```
lziest: why don't we call it Conn instead?

server/server.go
joshlf: We don't have to have it - it's not a correctness thing - but without it, we're likely to get lots of spurious log messages. Consider, for example, the following sequence of events:
1. Reader has an error (not EOF) reading from the connection, so it logs the error and closes the connection
2. Because the connection has been closed, the writer gets and error that they are trying to write to a closed connection. They log that too.

Thus, two error messages are generated for the same event. The second error message isn't that helpful, either, since it doesn't reflect the error that really caused the issue.

joshlf: To be fair, if there were a way to check whether a given error was due to an operation being performed on a closed connection, we could just use that, and I'll admit it'd be a lot cleaner and simpler. Unfortunately, I don't think anything like that exists (I've looked and haven't found anything).

server/internal/worker/client.go
joshlf: Because it doesn't have to be a network connection, and it particularly represents a client - not just any connection. It could be a client on the command line, for example, or something else entirely.

server/internal/worker/client.go
```diff
@@ -0,0 +1,383 @@
...
+	// and any future calls to those methods should also immediately return false.
+	//
+	// It must be safe to call Destroy multiple times; all calls after the first
+	// are no-ops.
+	Destroy()
+}
+
+// A ClientHandle is a handle on a pair of reader/writer goroutines that are
+// processing requests from a client.
+type ClientHandle struct {
```
lziest: I feel we are re-inventing connection duplex read/write. Is this something we borrowed from https://golang.org/pkg/net/#Conn?

server/internal/worker/client.go
lziest: a keyless client doesn't use connection to reach server? I think I missed something.

server/internal/worker/client.go
joshlf: This is unfortunately much more complicated than that. `net.Conn` allows you to perform multiple operations on a connection at a time. What we're doing here is abstracting the concept of receiving requests for work from a client, performing those requests, and sending back responses. The functionality here also includes out-of-band backpressure to avoid DoS vulnerabilities (see the long comment at the top of the file).

server/internal/worker/client.go
lziest: We are shifting the model? The server gets a bunch connections, and deal with them. Each connection comes from one client, one client may create several connections. So one client represents a bunch connections now?

joshlf: In practice it will usually be a network connection, yes. But this type represents something more structured than a network connection - it represents a connection to a client in particular - one who is submitting requests for work, and expects responses.

joshlf: No, one client still represents one connection.

server/server.go
lziest: agree

server/server.go
lziest: np

PR #176, opened 2017-07-18
Bootstrap an RPC server through the gokeyless protocol.

Bren2010:
- This is not currently merge-able. I'll have to rebase it on top of JoshLF's changes once they're worked through.
- The tests don't pass because I'm missing the travis.yaml bump (will be fixed with the above).

Opening now in case anyone wants to do preliminary review.


grittygrease:
Interesting feature, what's the motivation for this modification?

Bren2010:
Oh, that. I knew I was forgetting something.

Currently, when we want to add a new feature to keyless, we have to make manual changes at the protocol-level to support the new feature. This is bad because:
- We are open-sourcing nonsense bloat. None of the open-source code uses OpGetCertificate, OpSeal, OpUnseal, but that code is here because... here is where the protocol is specified.
- We have to coordinate between multiple repos. So things that would be one coherent PR, turn into multiple PRs. Then when one PR is merged and you update the vendor in others, you risk pulling in unintentional changes.

Because of that, I think we should only manually implement the protocol for features that are truly open-source and features that need to be compatible with C, and otherwise use RPC.

I believe OpSeal / OpUnseal need to work with C, but I would like to use this to get rid of OpGetCertificate.

grittygrease:
This sounds sane, I support it.

Bren2010:
Rebased on master.

joshlf:
Ah this looks awesome! I'm such a fan of this idea. Reviewing now...

conn/conn.go
```diff
@@ -306,3 +309,64 @@ func (c *Conn) Ping(data []byte) error {
 		return fmt.Errorf("ping: got unexpected response opcode: %v", result.Opcode)
 	}
 }
+
+// RPC returns an RPC client which uses the connection. Closing the client will
+// cleanup any spawned goroutines, but will not close the underlying connection.
```
joshlf: "Closing the client" means closing the `*rpc.Client` that is returned? I'd clarify the comment.

conn/conn.go
```diff
@@ -306,3 +309,64 @@ func (c *Conn) Ping(data []byte) error {
...
+func (cc *clientCodec) ReadResponseHeader(res *rpc.Response) error {
+	// gob decoders are stateful but we encode statelessly, so we need to reset
+	// the decoder after every full read.
+	cc.dec = gob.NewDecoder(cc.pr)
+	return cc.dec.Decode(res)
+}
+
+func (cc *clientCodec) ReadResponseBody(body interface{}) error {
+	return cc.dec.Decode(body)
+}
```
joshlf: So if I'm reading this right, this isn't safe for concurrent access, and `ReadResponseHeader` must be called before `ReadResponseBody` or else `cc.dec` will be nil?

server/server.go
```diff
@@ -194,6 +200,11 @@ func (s *Server) SetSealer(sealer Sealer) {
 	s.sealer = sealer
 }
 
+// RegisterRPC publishes in the server the methods on rcvr.
+func (s *Server) RegisterRPC(rcvr interface{}) error {
+	return s.dispatcher.Register(rcvr)
+}
```
joshlf: Can you explain how this method is used? I'm not sure that I understand it. Who calls this method, and how are the registered methods called?

server/server.go
```diff
@@ -293,6 +304,19 @@ func (w *otherWorker) Do(job interface{}) interface{} {
...
+		err := w.s.dispatcher.ServeRequest(codec)
+		if err != nil {
+			log.Errorf("Worker %v: ServeRPC: %v", w.name, err)
+			w.s.stats.logInvalid(requestBegin)
+			return makeErrResponse(pkt.ID, protocol.ErrInternal)
+		}
+
+		w.s.stats.logRequestDuration(requestBegin)
+		return makeRespondResponse(pkt.ID, codec.response)
+
```
joshlf: Should we be concerned about latency here? If this runs in the normal worker pool, then blocking RPC requests could hang up the server's ability to do non-ECDSA operations. Are we expecting the RPC server to always be running on localhost or something?

conn/conn.go
Bren2010: rpc.Client ensures that the codec is used properly. There's one goroutine that calls "ReadRespHeader", "ReadRespBody" in lockstep.

server/server.go
Bren2010: See tests/client_test.go:293, tests/common_test:192. RegisterRPC is called where SetGetCert, SetSealer would be. A client calls RPC methods by creating an RPC client with Conn.RPC(), and using "*rpc.Client.Call(...)".

server/server.go
Bren2010: It has the same latency concerns as a GetCert or a Sealer.

conn/conn.go
joshlf: Oh, I see. That's guaranteed by the `rpc` package's behavior?

server/server.go
joshlf: Sorry, my point is that all non-ECDSA work is currently done in a single worker pool, so a sufficient number of these requests, if they blocked for a meaningful amount of time, could block all other non-ECDSA operations on the server. So my question is: are we expecting these RPC servers to be run on the same machine as the Keyless server, or might they be run elsewhere across the network?

server/server.go
joshlf: Right, but my point is, would this be called by somebody setting up a server (e.g., what we've got in `cmd/gokeyless`), or by an actual client (e.g., in the standard Cloudflare edge/origin server running Keyless setup, called by the Cloudflare edge)?

conn/conn.go
Bren2010: Yes, it is. It's the only direct consumer of this object.

joshlf: OK sounds good.

server/server.go
Bren2010: The RPC server isn't a separate thing, that the keyless server is proxying to. The keyless server *is* the RPC server.

server/server.go
Bren2010: I'm not sure I understand the question. RegisterRPC is called by the server, with APIs that it would like to expose to clients throug the keyless protocol.

server/server.go
joshlf: Hmm I think I'm not understanding how the architecture works here. Let's say that I want to run a Keyless server with some extra plugins that I wrote - what do I do?

Bren2010: When you're configuring the server, where you'd call SetGetCert / SetSeal, you'd also call RegisterRPC(Obj{}). This would allow clients to call methods on Obj (and these methods can modify the fields of Obj) through the keyless protocol.

joshlf: OH OK I think I understand now. Let me know if I've got this right: the client sends a message with the opcode OpRPC and whose payload is an encoded set of RPC arguments that will be interpreted by the `rpc` package. The server extracts the payload and passes it to the dispatcher, which in turn dispatches to a dynamically-registered RPC handler?

Bren2010: Yes, that's right.

joshlf:
Can you add some documentation on the architecture here (as in, how RPCs flow from the client to the server to the plugin and back)? It wasn't obvious to me, at least.

joshlf:
Since this constitutes a protocol change, do we need to also update the C implementation and the protocol documentation? Also, do we have some story on backwards compatibility? As in, how should a client that knows about OpRPC handle a server that is older and responds to OpRPC messages with an error?

lziest:
mostly OpRPC would be internal use, the C implementation is too far behind to be relevant at this point.

joshlf:
> the C implementation is too far behind to be relevant at this point.

In that case, we should probably update the README in the C repository to clarify that the implementation is deprecated and out of date, point at this repository, and make sure that the README in this repository is up-to-date on the current protocol.

Bren2010:
I don't want to change the C repo, and we already say in the gokeyless readme that this implementation is an upgrade from the C implementation.

joshlf:
I still think it's worth mentioning in the C repo, but that's a separate issue - I'll file an issue over there.

joshlf:
Other than that, this looks good to me. Let's wait for @lziest 's signoff, though.

lziest:
oh, lgtm

PR #179, opened 2017-08-29
Fix goroutine leak in server.

Bren2010:
Reading a TLS connection with len(b)==0 is a noop, but if IsAlive never returns false, then many goroutines never go away.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -21,6 +21,7 @@ import (
 	"regexp"
 	"runtime"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/cloudflare/cfssl/log"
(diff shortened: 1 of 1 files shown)


lziest:
can you explain it a bit? So it can be closed exactly once. How does it fix the leak?

Bren2010:
(Currently) conn.IsAlive tries to read from the connection with an empty buffer, hoping that it will do some network-level magic and return an error if the connection is closed. If it does return false, the healtcheck goroutine [here](https://github.com/cloudflare/gokeyless/blob/master/server/internal/client/client.go#L208) will clean up the goroutines that have been spawned to serve the connection.

But crypto/tls cuts off the network-level magic [[source](https://golang.org/src/crypto/tls/conn.go?s=31940:31988#L1106)], so IsAlive always returns true and these goroutines never leave.

server/server.go
```diff
@@ -575,20 +580,20 @@ func (c *conn) SubmitResult(result interface{}) (ok bool) {
...
 	}
 	return true
 }
 
 func (c *conn) IsAlive() bool {
-	var buf [0]byte
-	_, err := c.conn.Read(buf[:])
-	return err == nil
+	return atomic.LoadUint32(&c.closed) == 0
 }
```
joshlf: Checking `c.closed` shouldn't replace the existing check, but rather should be performed in addition - `IsAlive` should return false if either `err != nil` OR if `c.closed` is 0. The reason for this is that the connection can become closed without being explicitly closed by our code - if the other side closes the connection.

Bren2010: Reading with an empty buffer will never return an error.

lziest:
I'm inclined to this change, as I am not entirely certain about Go's backward/forward compatibility now.

server/server.go
joshlf: Oh, now I see what you mean about `tls.Conn`'s `Read` method. OK.

joshlf: But I agree with @lziest - we should probably include it anyway in case the behavior changes in the future. Can't hurt.

Bren2010: If we are tracking errors manually, I don't believe there is any point in doing both.

PR #180, opened 2017-09-05
Brendan/retry accept

Bren2010:
- Commit 1 essentially duplicates the same logic from net/http. If Accept fails for any reason, then the current code tries to shutdown the server, but really just hangs everything up because...
- Commit 2 stops bothering to destroy background workers. BackgroundPool.Destroy blocks forever, waiting for BackgroundPool.worker to exit (which it never will because nothing is consuming randoms). This stops the server from ever completing shutdown.

### DIFF
--- a/server/internal/ecdsa/ecdsa.go
+++ b/server/internal/ecdsa/ecdsa.go
@@ -1,6 +1,7 @@
 package ecdsa
 
 import (
+	"context"
 	"crypto"
 	"crypto/aes"
 	"crypto/cipher"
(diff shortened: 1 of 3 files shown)


joshlf:
Would it be possible to fix the hanging issue by closing the channel? Ie, add a `Close` method to `ecdsa.SyncRandBuffer`? I'd prefer to actually fix this if we can rather than just leak resources.

Bren2010:
I used a context, because it feels more appropriate (and also because it's funny to me).

joshlf:
Awesome, and you've verified that this doesn't hang like the old code did?

Bren2010:
Yes, I have

PR #183, opened 2017-09-12
Add a way to export utilization info.

Bren2010:


server/server.go
```diff
@@ -694,6 +714,8 @@ func (s *Server) ServeConfig(l net.Listener, cfg *ServeConfig) error {
 		bgpool.Destroy()
 		otherpool.Destroy()
 		ecdsapool.Destroy()
+		// Stop publishing utilization info.
+		close(utilCh)
```
joshlf: It'd be good if we could block until the goroutine quit (maybe a wait group). Currently the `bgpool`, `otherpool`, and `ecdsapool` `Destroy` methods block so this function overall won't return until everything's fully cleaned up.

PR #184, opened 2017-09-20
travis: update go versions

dvrkps:


joshlf:
Does this just test with the latest point release for each major version?

dvrkps:
@joshlf Yes, it does.

PR #186, opened 2017-09-28
Close doesn't lock the DoRead goroutine.

Bren2010:
Closing an idle connection causes a deadlock, because the `DoRead` goroutine isn't reading anything from the connection (so isn't releasing the `readMtx`) and the `Close` goroutine won't move until it can acquire the `readMtx`.

### DIFF
--- a/conn/conn.go
+++ b/conn/conn.go
@@ -114,8 +114,8 @@ type Conn struct {
 
 	opTimeout time.Duration
 
-	// If you're going to acquire all three, always acquire in the following order
-	// to avoid deadlock: readMtx, writeMtx, mapMtx.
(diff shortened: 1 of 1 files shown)


joshlf:
I assume you're OK with the fact that this will make it so that `Close` can cause outstanding calls to `DoRead` to return `io.EOF`?

Bren2010:
Why would that be a problem? `DoRead` is constantly being called, and blocking for minutes at a time. That's why the deadlock happens.

joshlf:
Yeah, I know. I don't think it's a problem, I just wanted to make sure you knew that that was going to be the new behavior. Honestly I think that it makes the most sense.

PR #192, opened 2017-10-25
server: Add RSA precomputation and multiprime benchmarks

joshlf:
Most of the time, RSA private keys should already include cached precomputed values. This PR adds a check that precomputes them in case they haven't been already. It's mostly a hedge against client code forgetting to precompute, which should be uncommon since most of the standard library precomputes under the hood when generating private keys (either from scratch or when parsing files).

### DIFF
--- a/server/microbench_test.go
+++ b/server/microbench_test.go
@@ -113,7 +113,7 @@ func benchRandParallelFor(b *testing.B, op func(rnd io.Reader)) {
 
 // benchRandForSignRSA is an RSA signature-specific wrapper around benchRandFor.
 func benchRandForSignRSA(b *testing.B, params params.RSASignParams) {
-	key, payload := prepareRSASigner(b, 2048, params.PayloadSize)
+	key, payload := prepareRSASigner(b, 2, 2048, params.PayloadSize, true)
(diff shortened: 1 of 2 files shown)


lziest:
lgtm

joshlf:
I'll ask for an "official" review then :)

tmthrgd:
`(*rsa.PrivateKey).Precompute()` isn't thread-safe, so this pull request will cause a race condition. It could also lead to pretty severe signing failures as the `decrypt(io.Reader, *PrivateKey, *big.Int) (*big.Int, error)` function in `crypto/rsa` will see partially calculated values. (This race condition would probably linger for a long time causing occasional obscure failures).

*(I don't actually work for Cloudflare but I find this product really interesting).*

joshlf:
That is a very good point!

I don't see a good way to do this that both preserves performance, is correct, and doesn't add a lot of complexity. Given the marginal benefit, I think I'll kill the precomputation and just keep the benchmarks.

Thanks @tmthrgd !

joshlf:
OK I've removed the precomputation code, added a comment to `Keystore.Get` suggesting precomputation, and added benchmarks for precomputation and, while I was at it, multi-prime RSA.

@lziest or @Bren2010 , it's ready for another review.

PR #200, opened 2017-11-22
server: Don't print byte slices as strings in logs

joshlf:


joshlf:
Actually @lziest hold off on this. I think this is incomplete.

PR #205, opened 2017-12-06
cmd/gokeyless: Specify zone ID in CF API request

joshlf:


cmd/gokeyless/initialize.go
```diff
@@ -21,14 +21,16 @@ type v4apiError struct {
 type initAPIRequest struct {
 	Rqtype    string   `json:"request_type,omitempty"`
 	Hostnames []string `json:"hostnames,omitempty"`
+	Zones     []string `json:"zones,omitemtpy"`
```
lziest: use `zone_id` instead of `zones`

Issue #214, opened 2017-12-18
Fix TestRSADecrypt


joshlf:
https://github.com/cloudflare/gokeyless/commit/27a3a854c55e79cd15611230f7f96aeaf6139806 broke the `TestRSADecrypt` test; see [on Travis](https://travis-ci.org/cloudflare/gokeyless/jobs/318275474).

joshlf:
Closed in #215.

PR #215, opened 2018-01-19
client: Implement RSA decryption.

Bren2010:
This fixes the tests (closing #214). However, there are no broken use-cases IRL that I can think of...

### DIFF
--- a/client/keys.go
+++ b/client/keys.go
@@ -173,20 +173,60 @@ func (key *Decrypter) Decrypt(rand io.Reader, msg []byte, opts crypto.DecrypterO
 	}
 
 	if ok {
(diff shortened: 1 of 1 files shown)


client/keys.go
```diff
@@ -173,20 +173,60 @@ func (key *Decrypter) Decrypt(rand io.Reader, msg []byte, opts crypto.DecrypterO
 	}
 
 	if ok {
-		// If opts.SessionKeyLen is set, we must perform a variation of
```
lziest: shall we keep this comment?

Bren2010: I don't think it's particularly helpful.

PR #219, opened 2018-01-22
Don't measure key load time as part of execution time metric

### LINKED ISSUE #211: Don't measure key load time as part of execution time metric
Currently, the metric for execution time that we expose includes the time to load a key, but this could add significant latency, and that's not what we're trying to measure. Instead, we should only measure starting after the key has been loaded.

joshlf:
- For requests that involve loading a private key, don't include the time to load that key in the measured duration of the request execution.
- Make `logInvalid` not measure request execution duration.
- Don't log total request duration (`logRequestTotalDuration`) when there is an error in executing the request.

Closes #211

### DIFF
--- a/server/metrics.go
+++ b/server/metrics.go
@@ -88,9 +88,8 @@ func newStatistics() *statistics {
 }
 
 // logInvalid increments the error count and updates the error percentage.
-func (stats *statistics) logInvalid(opcode protocol.Op, requestBegin time.Time) {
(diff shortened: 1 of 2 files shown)


Bren2010:
Why don't we want to measure key load time?

joshlf:
We do, and we already have a metric for it. What this PR does is it makes it so that the metric that measures the time to execute a request doesn't include the key load time - it only measures the time to execute the request after the key has already been loaded.

PR #221, opened 2018-01-24
protocol: Validate IP length when unmarshalling into Operation

### LINKED ISSUE #220: Validate that the ServerIP and ClientIP fields are IPv4 or IPv6 addresses

joshlf:
Closes #220

### DIFF
--- a/protocol/protocol.go
+++ b/protocol/protocol.go
@@ -519,11 +519,19 @@ func (o *Operation) MarshalBinary() ([]byte, error) {
 }
 
 // UnmarshalBinary unmarshals a binary-encoded TLV list of items into o.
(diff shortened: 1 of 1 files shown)


protocol/protocol.go
```diff
@@ -552,9 +560,19 @@ func (o *Operation) UnmarshalBinary(body []byte) error {
 				copy(o.Digest[:], data)
 			}
 		case TagClientIP:
-			o.ClientIP = data
+			ip, err := validateIP(data)
+			if err == nil {
```
Bren2010: if err != nil {
  return err
}
o.IP = ip

joshlf: Better?

PR #222, opened 2018-01-26
Revert "server: Modify metrics to log number of RSA primes in certain…

Bren2010:
… cases"

This reverts commit b07bc9ccdd084e65b6eb07ca28b82756f2f5879e.

I don't think we intend to exercise this information, and it creates a DoS vector (possibly large number of labels).

### DIFF
--- a/server/metrics.go
+++ b/server/metrics.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"fmt"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


joshlf:
What's the DoS vector? The only person that can cause the number of labels to be high is the person running the server since they're the one that loads the private keys.

Bren2010:
It's a problem in Cloudflare's use-case, where we serve use-provided keys.

lziest:
I am a little confused, I thought this logging is for our internal use initially... What would number of primes help us to know?

joshlf:
Superseded by #224.

PR #223, opened 2018-01-26
protocol.Operation's stringer is defined on its pointer.

Bren2010:


protocol/protocol.go
```diff
@@ -366,7 +366,7 @@ type Operation struct {
 }
 
 func (o *Operation) String() string {
-	return fmt.Sprintf("[Opcode: %02x, SKI: %v, Digest: %02x, Client IP: %s, Server IP: %s, SNI: %s]",
+	return fmt.Sprintf("[Opcode: %s, SKI: %v, Digest: %02x, Client IP: %s, Server IP: %s, SNI: %s]",
```
joshlf: Shouldn't that be `%v`? `Opcode` implements `Stringer`.

PR #224, opened 2018-01-29
server: Fix measurements of the number of primes in an RSA key

joshlf:
- Don't warn about RSA keys that are not `*rsa.PrivateKey`.
- Impose a limit of 10 primes in order to avoid a DoS vector.


joshlf:
@Bren2010 Done; good call.

PR #225, opened 2018-02-15
Migrate to dep

cbroglie:
Moving the vendor folder to the top level addresses issues around mixed
dependency versions.


lziest:
I haven't used `dep` yet. One question, if I vendor 'gokeyless' lib in my code with `dep`. Would it vendor the nested vendor directory?

lziest:
If we don't create nested vendor lib, this LGTM

Bren2010:
What is the purpose of moving to dep? Moving vendor to the top-level breaks other people's ability to import packages at the top-level.

cbroglie:
As long as others importing this use `dep` as well, there shouldn't be any issues: https://github.com/golang/dep/blob/master/docs/FAQ.md#my-dependers-don-t-use-dep-yet-what-should-i-do

PR #226, opened 2018-02-22
Don't ask for user input by default

cbroglie:
The default behavior of the keyless server when the `server.pem` and `server-key.pem` files are not present is to ask the user for input to generate them. But if the server was launched as a daemon, it will just be stuck waiting for input, while appearing to be running.

With this change, the user will not be prompted for input unless the `-config-only` or `-manual-activation` flags are provided.

### DIFF
--- a/cmd/gokeyless/gokeyless.go
+++ b/cmd/gokeyless/gokeyless.go
@@ -95,8 +95,13 @@ func main() {
 		os.Exit(0)
 	}
 
+	// If we make it here we need to ask for user input, so we need to give up
+	// and log an error instead (in case the server is running as a daemon).
+	// Failing hard with an error message makes the problem obvious, whereas a
+	// daemon blocked waiting on input can be hard to debug.
 	if needNewCertAndKey() {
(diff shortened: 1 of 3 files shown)


cmd/gokeyless/gokeyless.go
```diff
@@ -95,8 +95,13 @@ func main() {
 		os.Exit(0)
 	}
 
+	// If we make it here we need to ask for user input, so we need to give up
+	// and log an error instead (in case the server is running as a daemon).
+	// Failing hard with an error message makes the problem obvious, whereas a
+	// daemon blocked waiting on input can be hard to debug.
 	if needNewCertAndKey() {
-		initializeServerCertAndKey()
```
lziest: so seems like it will not do auto-bootstrapping at all?

cbroglie: The intention is to only auto-bootstrap when the ‘-config-only’ flag is provided

lziest: sure, got it!

PR #230, opened 2018-05-04
Share a common pool of workers across listeners

cbroglie:
The pool will be created when the first listener is started, and will be
torn down after the last listener is closed.

And since we are going to share a pool of workers, we also needed to
refactor `Server` to share a config across listeners. Users who wish to
vary the config across listeners will need to configure multiple `Server`
instances.


lgarofalo:
LGTM

PR #231, opened 2018-05-14
Support configuration via file

cbroglie:


Makefile
```diff
@@ -13,7 +11,8 @@ PREFIX                       := usr/local
 INSTALL_BIN                  := $(DESTDIR)/$(PREFIX)/bin
 INIT_PREFIX                  := $(DESTDIR)/etc/init.d
 SYSTEMD_PREFIX               := $(DESTDIR)/lib/systemd/system
-CONFIG_PREFIX                := $(DESTDIR)/etc/keyless
+CONFIG_PATH                  := etc/keyless
```
mahrud: Shouldn't CONFIG_PATH also be relative to DESTDIR?

cmd/gokeyless/gokeyless.go
```diff
@@ -93,41 +212,69 @@ func main() {
...
 
 	// If we make it here we need to ask for user input, so we need to give up
 	// and log an error instead (in case the server is running as a daemon).
 	// Failing hard with an error message makes the problem obvious, whereas a
 	// daemon blocked waiting on input can be hard to debug.
 	if needNewCertAndKey() {
-		log.Error("the server cert/key need to be generated; run the server with either the -config-only or -manual-activation flag to generate the pair")
-		os.Exit(1)
+		if needInteractivePrompt() {
+			log.Error("the server cert/key need to be generated; set the hostname, zone_id, and origin_ca_key values in your config file, or run the server with either the -config-only or -manual-activation flag to generate the pair interactively")
```
Bren2010: log.Fatal?

cmd/gokeyless/gokeyless.go
```diff
@@ -93,41 +212,69 @@ func main() {
...
 			f.Close()
 		}
 	}
 
-	go func() { log.Critical(s.MetricsListenAndServe(metricsAddr)) }()
-	log.Fatal(s.ListenAndServe(net.JoinHostPort("", port)))
+	go func() {
+		log.Critical(s.MetricsListenAndServe(net.JoinHostPort("", strconv.Itoa(config.MetricsPort))))
+	}()
+	log.Fatal(s.ListenAndServe(net.JoinHostPort("", strconv.Itoa(config.Port))))
```
Bren2010: We imported strconv just for Itoa when we already have fmt?

cmd/gokeyless/gokeyless.go
```diff
@@ -22,55 +27,169 @@ const (
...
+type Config struct {
+	LogLevel int `yaml:"loglevel" mapstructure:"loglevel"`
+
+	Hostname     string `yaml:"hostname" mapstructure:"hostname"`
+	ZoneID       string `yaml:"zone_id" mapstructure:"zone_id"`
+	OriginCAKey  string `yaml:"origin_ca_key" mapstructure:"origin_ca_key"`
+	InitEndpoint string `yaml:"init_endpoint" mapstructure:"init_endpoint"`
+
+	CertFile   string `yaml:"cert" mapstructure:"cert"`
+	KeyFile    string `yaml:"key" mapstructure:"key"`
```
Bren2010: Where is the mapstructure field tag consumed?

cmd/gokeyless/gokeyless.go
Bren2010: The first lines of initializeServerCertAndKey are:
```
if needsInteractivePrompt() {
  interactivePrompt()
}
```
but we've cut it off here. Is initializeServerCertAndKey called somewhere else?

pkg/gokeyless.yaml
```diff
@@ -0,0 +1,24 @@
...
+# keys.
+cert: /etc/keyless/server.pem
+key: /etc/keyless/server-key.pem
+ca_cert: /etc/keyless/keyless_cacert.pem
+csr: /etc/keyless/server.csr
+# Optionally customize the listen ports.
+port: 2407
+metrics_port: 2406
+# Optionally write the PID to a file.
+pid_file:
```
Bren2010: Add a newline?

cmd/gokeyless/gokeyless.go
cbroglie: The [viper](https://github.com/spf13/viper) package I'm using to handle the config parsing uses mapstructure for parsing.

The `yaml` tags are not actually needed for parsing, but I included them for serializing with the config with the yaml package.

Makefile
mahrud: Then I don't understand why config_path is used instead of config_prefix in lines 65 and 77.

cmd/gokeyless/gokeyless.go
cbroglie: Yes, it's called when the `--config-only` flag is provided, which is handled earlier in the flow.

The idea is to only run ever run the interactive flow if that flag is provided, while also still attempting to auto bootstrap if we have all the input we need from the user via config.

Can switch to `log.Fatal` if it's preferred - the existing conventions are mixed.

pkg/gokeyless.yaml
cbroglie: 👍

Makefile
cbroglie: B/c those commands are referring to the file path after installation. The context is different.

mahrud: Wait, isn't the purpose of DESTDIR to, for instance, allow the user to install under /usr/local/etc/keyless? If not then never mind, I misunderstood this.

cmd/gokeyless/gokeyless.go
cbroglie: So we did. But the strconv package is already imported elsewhere in this project so there is no impact to binary size.

Makefile
cbroglie: DESTDIR is set directly above here

server/server.go
```diff
@@ -61,32 +61,45 @@ func NewDefaultKeystore() *DefaultKeystore {
 // into a crypto.Signer, which is stored in the Keystore.
 func NewKeystoreFromDir(dir string, LoadKey func([]byte) (crypto.Signer, error)) (Keystore, error) {
```
mahrud: Do we still need this?
I can have client/remote_test.go and tests/common_test.go also use AddFromDir, then there would be no reason to keep this around.

cbroglie: This project contains both a library and an application which uses the library. By leaving this in place we add the new functionality we need without needlessly breaking the library for anyone who was happily using the existing function.

PR #11, opened 2015-08-04
Add generic Keystore for server private keys.

jacobhaven:
This will allow for smarter lazy-loading implementations capable of loading keys based on incoming requests.

PR #232, opened 2018-05-31
[WIP] Add support for HSMs

cbroglie:


cmd/gokeyless/gokeyless.go
```diff
@@ -147,10 +148,12 @@ func initConfig() error {
...
 	for _, store := range config.PrivateKeyStores {
-		if (store.Dir == "" && store.File == "") ||
-			(store.Dir != "" && store.File != "") {
-			return fmt.Errorf("private key stores must define exactly one of the 'dir' or 'file' keys")
+		if (store.Dir != "" && store.File == "" && store.URI == "") ||
+			(store.Dir == "" && store.File != "" && store.URI == "") ||
+			(store.Dir == "" && store.File == "" && store.URI != "") {
+			continue
 		}
+		return fmt.Errorf("private key stores must define exactly one of the 'dir' or 'file' keys")
```
cbroglie: Please update this error message to account for the 'uri' option

server/pkcs11.go
```diff
@@ -0,0 +1,158 @@
...
+	pChar := "[a-zA-Z0-9-_.~%:\\[\\]@!\\$'\\(\\)\\*\\+,=&]"
+	pAttr := aChar+"+="+pChar+"+"
+	pClause := "("+pAttr+";)*("+pAttr+")"
+	qChar := "[a-zA-Z0-9-_.~%:\\[\\]@!\\$'\\(\\)\\*\\+,=/\\?\\|]"
+	qAttr := aChar+"+="+qChar+"+"
+	qClause := "("+qAttr+"&)*("+qAttr+")"
+	r, _ := regexp.Compile("^pkcs11:"+pClause+"(\\?"+qClause+")?$")
+
+	if ! r.MatchString(uri) {
+		log.Error("PKCS#11 URI is malformed: " + uri)
```
cbroglie: Instead of logging errors, the function signature should include an error return value, and you should return an error here. The calling code can decide about logging (think of this function like `strconv.Atoi`).

server/pkcs11.go
```diff
@@ -0,0 +1,158 @@
...
+			log.Warning("Unrecognized PKCS#11 URI Query: " + parts[0])
+		}
+	}
+
+	return pk11uri
+}
+
+// LoadPKCS11Key attempts to load a signer from a PKCS#11 URI.
+// See https://tools.ietf.org/html/rfc7512#section-2.3
+func LoadPKCS11Key(pk11uri PKCS11URI) (priv crypto.Signer, err error) {
```
cbroglie: Let's not use named return arguments here, since you're not actually using them. It's easy to introduce subtle bugs with the variable shadowing.

server/pkcs11.go
```diff
@@ -0,0 +1,158 @@
...
+			pk11uri.Id = []byte(value)
+		case "slot-manufacturer":
+			pk11uri.SlotManuf = value
+		case "slot-description":
+			pk11uri.SlotDesc = value
+		case "slot-id":
+			id, _ := strconv.ParseUint(value, 10, 32) // TODO what is the bit size
+			pk11uri.SlotId = uint(id)
+		default:
+			log.Warning("Unrecognized PKCS#11 URI Component: " + parts[0])
```
cbroglie: Instead of logging an error, let's return one here. It's best to fail loudly when introducing new logic as we can more easily catch problems.

server/pkcs11.go
```diff
@@ -0,0 +1,158 @@
...
+		case "pin-source":
+			pk11uri.PinSource = value
+		case "pin-value":
+			pk11uri.PinValue = value
+		case "module-name":
+			pk11uri.ModuleName = value
+		case "module-path":
+			pk11uri.ModulePath = value
+		default:
+			log.Warning("Unrecognized PKCS#11 URI Query: " + parts[0])
```
cbroglie: Same comment - return error instead of logging.

server/pkcs11.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/thalesignite/crypto11"
+	"github.com/cloudflare/cfssl/log"
+)
+
+// PKCS11URI defines a PKCS#11 URI: https://tools.ietf.org/html/rfc7512#section-2.3
+type PKCS11URI struct { // pkcs11:path-attr[?query-attr]
+	// path-attr: (; delimited)
+	Token  string //        token <- CK_TOKEN_INFO
+	Manuf  string // manufacturer <- CK_TOKEN_INFO
```
cbroglie: I believe the idiomatic Go thing would be to spell these fields out, i.e. `Manufacturer` instead of `Manuf`, `LibraryDescription` instead of `LibDesc`, etc.

server/pkcs11.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// query-attr: (& delimited)
+	PinSource string // pin-source
+	PinValue  string //  pin-value
+
+	ModuleName string // module-name
+	ModulePath string // module-path
+}
+
+func RFC7512Parser(uri string) PKCS11URI {
```
cbroglie: If you return `*PKCS11URI` instead of `PKCS11URI` it will save a few mem copies.

server/pkcs11_test.go
```diff
@@ -0,0 +1,28 @@
...
+pkcs11:id=0;token=b
+pkcs11:id=b?pin-value=1234
+pkcs11:id=b?pin-value=1234&module-path=/a/b.so
+pkcs11:id=b;token=b?pin-value=1234&module-path=/a/b.so
+pkcs11:token=Thalesn6000;id=%00?module-path=/opt/nfast/toolkits/pkcs11/libcknfast.so&pin-value=1234
+pkcs11:token=SoftHSM2;id=%00;slot-id=1359138056?module-path=/usr/lib64/libsofthsm2.so&pin-value=1234
+pkcs11:token=YubiKey%20PIV;id=%00;slot-id=0?module-path=/usr/lib64/libykcs11.so&pin-value=123456
+pkcs11:token=Gemalto;id=%04;slot-id=0?module-path=/usr/lib/libCryptoki2_64.so&pin-value=abcd`
+
+// TODO add false tests too
```
cbroglie: Yes, let's do that. I think you mentioned that you saw some pkcs#11 parser libraries which looked like personal projects; it's probably worth looking at those just to borrow their test cases.

tests/common_test.go
```diff
@@ -111,11 +119,22 @@ func init() {
 		log.Fatal(err)
 	}
 
-	keys, err := server.NewKeystoreFromDir("testdata", LoadKey)
-	if err != nil {
-		log.Fatal(err)
+	if os.Getenv("TESTHSM") == "" {
```
cbroglie: Does this need to be mutually exclusive? How about writing it as:
```go
keys, err := server.NewKeystoreFromDir("testdata", LoadKey)
if err != nil {
	log.Fatal(err)
}
if os.Getenv("TESTHSM") != "" {
	if err := keys.AddFromURI(params.RSAURI, LoadURI); err != nil {
		log.Fatal(err)
	}
	if err := keys.AddFromURI(params.ECDSAURI, LoadURI); err != nil {
		log.Fatal(err)
	}
}
s.SetKeystore(keys)
```

tests/testdata/ecdsa.key
```diff
@@ -1,5 +1,5 @@
 -----BEGIN EC PRIVATE KEY-----
-MHcCAQEEINL8ATxoL9YROkZ0KN4psqHHkfuXS7CnoKHTpQcJ1rW7oAoGCCqGSM49
-AwEHoUQDQgAE/zrRppSJW7b7ydzj9HZuZchDAqzJ6Q0w48AT51Gdn/teiwpjbHbR
-CdK9EneBjKOC7LNiacMz8tkixhI9ysBikg==
+MHcCAQEEINHLf686Kw39K94sKXC+oYNs45qVfnXbUBacgRDuCOHQoAoGCCqGSM49
```
cbroglie: I'm just curious, why did these certs get changed?

pkg/gokeyless.yaml
```diff
@@ -13,6 +13,7 @@ origin_ca_key:
 # Configure one or more private key directories.
 private_key_stores:
 - dir: /etc/keyless/keys
+- uri: pkcs11:token=SoftHSM2%20RSA%20Token;id=%03;slot-id=43989470?module-path=/usr/lib64/libsofthsm2.so&pin-value=1234
```
cbroglie: This is the default config file that gets installed with the deb/rpm. Test data does not belong here.

Makefile
```diff
@@ -83,3 +83,14 @@ $(RPM_PACKAGE): | $(INSTALL_BIN)/$(NAME) install-config
...
+
+SOFTHSM_TOKENS_PATH          := $(DESTDIR)/var/lib/softhsm/tokens
+SOFTHSM_CONFIG_PATH          := $(DESTDIR)/etc
+
+.PHONY: dev-softhsm
+dev-softhsm:
+	@mkdir -p  $(SOFTHSM_CONFIG_PATH)
+	@mkdir -p  $(SOFTHSM_TOKENS_PATH)
+	@chmod 700 $(SOFTHSM_TOKENS_PATH)
+	@cp -r pkg/testing-tokens/*      $(SOFTHSM_TOKENS_PATH)
```
cbroglie: The `pkg` folder for files which are included in the final deb/rpm packages, not test data. You should use the `testdata` folder for that. Using `$(DESTDIR)` doesn't make sense here either, as that's also for the deb/rpm packaging. The logic you have here makes sense, it just needs to use different folders rather than hijacking the deb/rpm build.

tests/testdata/ecdsa.key
mahrud: The previous key/certs would not work with SoftHSM2!

tests/common_test.go
mahrud: One returns `Keystore` and the other returns `DefaultKeystore`, so this seemed the easiest.

cbroglie: Well how about this then:
```go
keys := server.NewDefaultKeystore()
if err := keys.AddFromDir("testdata", LoadKey); err != nil {
	log.Fatal(err)
}
if os.Getenv("TESTHSM") != "" {
	if err := keys.AddFromURI(params.RSAURI, LoadURI); err != nil {
		log.Fatal(err)
	}
	if err := keys.AddFromURI(params.ECDSAURI, LoadURI); err != nil {
		log.Fatal(err)
	}
}
s.SetKeystore(keys)
```

Makefile
mahrud: But look in `install-config`, you're copying testing keys in the same place. How is this different? I'm just following that example.

cbroglie: That's exactly my point - that target is used for building the deb/rpm packages.

pkg/gokeyless.yaml
mahrud: How do you propose to test using a sample SoftHSM key? I added this here since there are testing keys in the `keys` directory as well. Even if SoftHSM isn't install, this doesn't lead to a fatal error.

server/pkcs11.go
mahrud: I changed `Id` to `ID` but as for others, I'd rather stick to the RFC convention here.

pkg/gokeyless.yaml
cbroglie: Tests should have their own test config file. I don't know why those test keys were added before, but they don't belong either. Having this random test config installed when customers install the package makes no sense at all.

tests/common_test.go
mahrud: What goal does this serve? The keys in those URI's are the exact keys in that directory.

server/pkcs11.go
```diff
@@ -0,0 +1,165 @@
...
+	PinValue  string //  pin-value
+
+	ModuleName string // module-name
+	ModulePath string // module-path
+
+	// Vendor specific query attributes:
+	MaxSessions int
+}
+
+func RFC7512Parser(uri string) (*PKCS11URI, error) {
```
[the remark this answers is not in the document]
Bren2010: My brain is shutting down trying to read the code below, please try to make it more readable. One approach: Break the function into chunks of 10-20 lines that accomplish a single coherent task, and put an inline comment above each chunk saying what that task is. Maybe just saying what some key variables represent and a brief description of the format is enough.

server/pkcs11.go
```diff
@@ -0,0 +1,165 @@
...
+			pk11uri.MaxSessions = int(maxSessions)
+		default:
+			return nil, fmt.Errorf("Unrecognized PKCS#11 URI Query: %s", parts[0])
+		}
+	}
+
+	return &pk11uri, nil
+}
+
+// LoadPKCS11Key attempts to load a signer from a PKCS#11 URI.
```
Bren2010: This is just a restatement of the function signature. Does this talk to the HSM? Does it have any notable error cases?

server/pkcs11.go
```diff
@@ -0,0 +1,165 @@
...
+		Path:        pk11uri.ModulePath,
+		TokenSerial: pk11uri.Serial,
+		TokenLabel:  pk11uri.Token,
+		Pin:         pk11uri.PinValue,
+		MaxSessions: pk11uri.MaxSessions,
+	}
+
+	_, err := crypto11.Configure(config)
+	if err != nil {
+		log.Warning(err)
```
Bren2010: Logging outside of func main is why the Alamo happened. Remember the Alamo.

server/pkcs11_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func TestParser(t *testing.T) {
+	tests := strings.Split(samples, "\n")
+	for _, test := range tests {
+		if _, err := RFC7512Parser(test); err != nil {
+			t.Fail()
```
Bren2010: t.Fail(err)

internal/test/params/params.go
```diff
@@ -51,3 +50,33 @@ var (
...
+	// testdata/tokens/d6a8ab57-d5c5-aaf0-70b6-d01595c28127 -> /var/lib/softhsm/tokens/
+	ECDSAURI = "pkcs11:token=SoftHSM2%20EC%20Token;id=%02;slot-id=1400733853?module-path=/usr/lib64/libsofthsm2.so&pin-value=12345"
+)
+
+// HSMSignParams represents a set of parameters to a HSM signing operation.
+type HSMSignParams struct {
+	Opcode      protocol.Op       // The Keyless protocol opcode for this operation
+	URI         string            // The PKCS#11 URI of the key to be used for signing
+	Opts        crypto.SignerOpts // Options to the signing function
+	PayloadSize int               // The size of the payload to be signed
```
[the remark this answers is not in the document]
mahrud: Good idea to simplify but I followed suit here, see `RSASignParams` and `ECDSASignParams` few lines above.

server/microbench_test.go
```diff
@@ -52,6 +54,29 @@ func benchSignParallel(b *testing.B, key crypto.Signer, rnd io.Reader, payload [
 	wg.Wait()
 }
 
+func benchSignConcurrency(b *testing.B, key crypto.Signer, rnd io.Reader, payload []byte, opts crypto.SignerOpts) {
```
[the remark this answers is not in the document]
mahrud: This isn't designed to test gokeyless, it's designed to test `crypto11` and the HSM module/device.

server/pkcs11.go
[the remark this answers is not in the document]
mahrud: Because I specifically want people to go read the RFC before reading this code. My brain also shuts down when reading the RFC. I'm sorry but it's not my fault that the authors decided to go with a scheme that has no parser anywhere. See for yourself: https://tools.ietf.org/html/rfc7512#section-2.3

I would add those comments but even I don't know what half of those variables represent!

server/microbench_test.go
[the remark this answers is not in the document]
Bren2010: Then shouldn't it be in crypto11?

server/pkcs11.go
```diff
@@ -0,0 +1,165 @@
...
+	"net/url"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/cbroglie/crypto11"
+	"github.com/cloudflare/cfssl/log"
+)
+
+// PKCS11URI defines a PKCS#11 URI: https://tools.ietf.org/html/rfc7512#section-2.3
```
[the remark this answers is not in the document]
mahrud: That's why I put a link to the RFC up there, because all of these are defined and used in there and if there is any confusion about what parameter X does you should refer to that RFC.

server/microbench_test.go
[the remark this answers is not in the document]
mahrud: That's a good point, but I think it's worth it to compare signing performance with and without HSM.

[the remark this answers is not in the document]
Bren2010: I believe that's what the 'parallel' benchmarks are measuring

server/microbench_test.go
```diff
@@ -184,6 +209,34 @@ func prepareECDSASigner(b *testing.B, curve elliptic.Curve, payloadsize int) (ke
...
+}
+
+func benchHSMSignParallel(b *testing.B, params params.HSMSignParams) {
+	pk11uri, _ := RFC7512Parser(params.URI)
+	key, payload := prepareHSMSigner(b, pk11uri, params.PayloadSize)
+	benchSignParallel(b, key, rand.Reader, payload[:], params.Opts)
+}
+
+func benchHSMSignConcurrency(b *testing.B, params params.HSMSignParams) {
+	pk11uri, _ := RFC7512Parser(params.URI)
```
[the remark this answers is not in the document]
Bren2010: Ok, and why?

server/pkcs11.go
```diff
@@ -0,0 +1,172 @@
...
+	ModulePath string // module-path
+
+	// Vendor specific query attributes:
+	MaxSessions int // max-sessions
+}
+
+// This function parses PKCS#11 URIs defined in RFC 7512 into PKCS11URI objects.
+// An error is returned if the input string does not appear to follow the rules
+// or if there are unrecognized components.
+func RFC7512Parser(uri string) (*PKCS11URI, error) {
```
Bren2010: The function name is bad, and the function docs should start with "FunctionName parses...", and the function docs are currently a restatement of the function signature.

server/pkcs11.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, fmt.Errorf("PKCS#11 URI is malformed: %s", uri)
+	}
+
+	var pk11uri PKCS11URI
+	var pAttrs []string
+	var qAttrs []string
+	var parts []string
+	var value string
+
+	// Then we parse the URI string
```
Bren2010: Should be a full sentence(ie, has a period), and isn't sufficiently descriptive

server/pkcs11.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+
+	return &pk11uri, nil
+}
+
+// LoadPKCS11Key attempts to load a Signer given a PKCS11URI object identifier.
+// An error is returned if the crypto11 module cannot find the module, token,
+// or the specified object.
+func LoadPKCS11Key(pk11uri *PKCS11URI) (crypto.Signer, error) {
```
Bren2010: The function docs are currently still a restatement of the function signature.

server/pkcs11_test.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	tests = strings.Split(negatives, "\n")
+	for _, uri := range tests {
+		if _, err := RFC7512Parser(uri); err == nil {
+			t.Fatalf("PKCS#11 Parser failed to detect a wrong URI: %s", uri)
+		}
+	}
+}
+
+func TestHSMSignConcurrencyRSASHA512(t *testing.T) {
```
Bren2010: What is the purpose of this test? Are there others like it?

tests/testdata/ecdsa.key
Bren2010: Why not?

server/pkcs11.go
mahrud: I think the function signature provides enough information about this function. I added a sentence for the error outputs as you asked.

server/pkcs11_test.go
mahrud: This is similar to the benchmark function with "Concurrency" in its name, except the number of concurrent goroutines is fixed here.

tests/testdata/ecdsa.key
mahrud: I don't know. Try it!

server/microbench_test.go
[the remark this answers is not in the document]
mahrud: For instance, we could add a test that generates a key on an HSM, stores the identifying attributes in a `PKCS11URI` object, and passes is to `prepareHSMSigner` to use. We would not want to encode the `PKCS11URI` into a string URI, then pass it here to be parsed again. Makes sense?

server/microbench_test.go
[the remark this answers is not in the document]
mahrud: No, they're different. The parallel benchmarks didn't reveal the bug in `crypto11`.

tests/testdata/ecdsa.key
Bren2010: Sounds like a bug, or at least non-linear behavior that will trip us up in the future. I'd recommend looking into it.

server/microbench_test.go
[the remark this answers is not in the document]
Bren2010: Ok, so put it in crypto11. It doesn't belong here.

server/pkcs11.go
Bren2010: "LoadPKCS11Key attempts to load a signer"
What does 'load' mean? Is it a symbolic operation, are there TCP connections made, are there syscalls?

server/pkcs11_test.go
Bren2010: Rephrased: What are you testing? We already established that the concurrency benchmarks should be removed.

mahrud: I'll wait @cbroglie to explain this when he's back. He wanted these tests.

server/pkcs11.go
```diff
@@ -0,0 +1,182 @@
...
+	ModulePath string // module-path
+
+	// Vendor specific query attributes:
+	MaxSessions int // max-sessions
+}
+
+// PKCS11Parser decodes a PKCS#11 URI as defined in RFC 7512 into an PKCS11URI object.
+// An error is returned if the input string does not appear to follow the rules
+// or if there are unrecognized path or query attributes.
+func PKCS11Parser(uri string) (*PKCS11URI, error) {
```
Bren2010: The function name should be declarative. I would prefer 'ParseURI' and we change PKCS11URI to just URI.

mahrud: I disagree, unless you want me to remove this code and make a `rfc7512` package, then `rfc7512.ParseURI` and `rfc7512.URI` would be appropriate.

server/pkcs11.go
mahrud: We're not doing any of that. We're simply calling a function in `crypto11` to load a key from a PKCS11 module and give it to us as a `crypto.Signer` object. I don't think the function name could be more declarative.

internal/rfc7512/rfc7512.go
```diff
@@ -0,0 +1,182 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/cbroglie/crypto11"
+)
+
+// A PKCS11URI represents a PKCS#11 object stored on a PKCS#11 token and
+// information for accessing it. The general form represented is:
+//
+//	pkcs11:path-attr[?query-attr]
```
Bren2010: What does this mean? More examples of copious documentation of unfamiliar formats:
- https://golang.org/pkg/text/template/
- https://golang.org/src/time/time.go#L328
- https://golang.org/pkg/regexp/syntax/

internal/rfc7512/rfc7512.go
```diff
@@ -0,0 +1,182 @@
+package rfc7512
```
Bren2010: Would `package hsm` work? And package docs

```
// Package hsm implements blah.
//
// More context if you want to give it. For more information, see this and that standard.
```

Bren2010: Could possibly also change PKCS11URI -> URI, LoadPKCS11Key -> LoadKey

internal/rfc7512/rfc7512.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}
+
+	return &pk11uri, nil
+}
+
+// LoadPKCS11Key attempts to load a Signer given a PKCS11URI object identifier.
+// An error is returned if the crypto11 module cannot find the module, token,
+// or the specified object.
+func LoadPKCS11Key(pk11uri *PKCS11URI) (crypto.Signer, error) {
```
Bren2010: > We're not doing any of that. We're simply calling a function in crypto11 to __load a key from a PKCS11 module__ and give it to us as a crypto.Signer object. I don't think the function name could be more declarative.

Please expand on the bolded section. "module, token, or the specified object" -- I don't know what any of these words mean.

tests/common_test.go
```diff
@@ -82,6 +85,12 @@ func LoadKey(in []byte) (priv crypto.Signer, err error) {
 	return derhelpers.ParsePrivateKeyDER(in)
 }
 
+// LoadURI attempts to load a private key from SOFTHSM.
+func LoadURI(uri string) (priv crypto.Signer, err error) {
+	pk11uri, _ := rfc7512.ParsePKCS11URI(uri)
+	return rfc7512.LoadPKCS11Key(pk11uri)
+}
```
Bren2010: This exact function was already implemented somewhere else. Dedup?

internal/rfc7512/rfc7512.go
mahrud: There already exists copious documentation of this unfamiliar format:
```
// For more information read: https://tools.ietf.org/html/rfc7512#section-2.3
```

Bren2010: Try re-writing and summarizing the relevant parts in your own words.

tests/common_test.go
mahrud: I think there was a decision made before defining `LoadKey` in `cmd/gokeyless/gokeyless.go` and passing it as a parameter, so I followed the same for `LoadURI`. We can discuss this and move them both to `server/` so they are importable.

mahrud: I made a change to fail if an error is returned and Github thought this is finished. It's not. I'll discuss it with Chris today.

internal/rfc7512/rfc7512.go
mahrud: I added package docs, but this isn't really an `hsm` package. Maybe `pkcs11uri` or `pkcs11-uri` but I honestly prefer `rfc7512`.
For struct and function names I'm following `crypto11`'s example for the most part.

Gopkg.lock
```diff
@@ -7,6 +7,12 @@
   packages = ["quantile"]
   revision = "4c0e84591b9aa9e6dcfdf3e020114cd81f89d5f9"
 
+[[projects]]
+  branch = "session-handles"
+  name = "github.com/cbroglie/crypto11"
```
mahrud: We can revert back to using `github.com/thalesignite/crypto11` since they merged your PR.

server/pkcs11_test.go
cbroglie: I wanted these tests b/c I had concerns about crypto11's handling of lots of session handles, and the tests did identify some problems which we fixed.

But I had completely forgotten about the fact that gokeyless uses a limited pool of workers rather than each new connection getting its own goroutine, so my concerns about those crypto11 issues affecting gokeyless turned out to be unfounded. @Bren2010 is correct that the test can be removed.

tests/testdata/ecdsa.key
Bren2010: @mahrud What was the issue here?

mahrud: I'm not sure and I can't replicate it anymore. It might have been me doing something incorrectly.
I've reverted the changes though, the keys haven't changed. I just moved them from `pkg` into `tests/testdata` to avoid installing testing keys.

internal/rfc7512/rfc7512.go
Bren2010: Stop 👏deferring 👏to 👏the 👏standards. Think of it like that book, The Giver: You read the standards so we don't have to.

Reference on names: https://blog.golang.org/package-names. This package lets you parse HSM URIs, it lets you 'connect' to an HSM. If this isn't an HSM package, what is a simple noun that describes it? The package docs are non-sensical; it'd be good to move the stuff in LoadPKCS11Signer's docs up here, with even more elaboration.

internal/rfc7512/rfc7512.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/ThalesIgnite/crypto11"
+)
+
+// PKCS11URI contains the information for accessing a PKCS #11 object, such as
+// a public key, private key, or a certificate. The general form represented is:
+//
+//	pkcs11:path-attr[?query-attr]
+//
+// Path and query attributes are delimited by ';' and '&', respectively.
+// For more information read: https://tools.ietf.org/html/rfc7512#section-2.3
```
Bren2010: Can you describe this better?

internal/rfc7512/rfc7512.go
```diff
@@ -0,0 +1,196 @@
...
+	aChar := "[a-z-_]"
+	pChar := "[a-zA-Z0-9-_.~%:\\[\\]@!\\$'\\(\\)\\*\\+,=&]"
+	pAttr := aChar + "+=" + pChar + "+"
+	pClause := "(" + pAttr + ";)*(" + pAttr + ")"
+	qChar := "[a-zA-Z0-9-_.~%:\\[\\]@!\\$'\\(\\)\\*\\+,=/\\?\\|]"
+	qAttr := aChar + "+=" + qChar + "+"
+	qClause := "(" + qAttr + "&)*(" + qAttr + ")"
+	r, _ := regexp.Compile("^pkcs11:" + pClause + "(\\?" + qClause + ")?$")
+
+	if !r.MatchString(uri) {
```
Bren2010: I believe it's possible to parse and validate this format without regex. Something like `url.ParseQuery`. Using that as a building block might make things a lot simpler.

internal/rfc7512/rfc7512.go
```diff
@@ -0,0 +1,196 @@
...
+//	Token:	use Serial or Token to specify the PKCS #11 token.
+//	Slot:	use SlotID, ID, or Object to specify the PKCS #11 key pair.
+//
+// For certain modules, a query attribute max-sessions is required in order to
+// prevent openning too many sessions to the module. Certain additional attributes,
+// such as pin-value, may be necessary depending on the situation. Refer to the
+// documentation for your PKCS #11 module for more details.
+//
+// An error is returned if the crypto11 module cannot find the module, token,
+// or the specified object.
```
Bren2010: I still don't really know what this function is doing. Try connecting the HSM-related terminology to other things I might know about. Is a module path a unix socket? Is a token a password?

internal/rfc7512/rfc7512.go
cbroglie: I much prefer @Bren2010's suggested naming conventions, and agree that your fellow programmers would be better suited with a short layman's description instead of just referencing the RFC. The Giver analogy is excellent.

That said, this is an internal package with pretty limited scope, so I'm not too stressed either way.

internal/rfc7512/rfc7512.go
mahrud: You need to know what is PKCS #11. I don't think this package is the right place to document that. You can read the documentation on the `pkcs11` package for that.

internal/rfc7512/rfc7512.go
mahrud: I think that's a good idea, but there are edge cases that will make it just as complicated. For instance path attributes can contain "&".

cbroglie:
Closing in favor of #235

PR #234, opened 2018-06-28
Differentiate key not found from other internal errors

cbroglie:
This change updates the `Keystore` interface so that callers can differentiate between keys not found, and other possible internal errors.

A new metric, `keyless_requests_internal_error`, was added to observe these errors. It's analogous to HTTP 5XX errors, whereas the existing `keyless_requests_invalid` metric represents HTTP 4XX errors. Other redundant metrics were consolidated without loss of fidelity:
- `keyless_requests` can be derived from `keyless_request_opcodes`
- `keyless_request_execution_duration` can be derived from `keyless_request_exec_duration_per_opcode`
- `keyless_request_total_duration` can be derived from `keyless_request_total_duration_per_opcode`

And the `rsa_primes` label was removed as it always has the value 2.


server/server.go
```diff
@@ -129,12 +129,11 @@ func (keys *DefaultKeystore) Get(op *protocol.Operation) (crypto.Signer, bool) {
 		priv, found := keys.skis[ski]
 		if found {
 			log.Infof("fetch key with SKI: %s", ski)
-			return priv, found
+			return priv, nil
 		}
```
Bren2010: Wouldn't you do `return nil, nil` for case `!found`? If so, I'd also make the last error more explicit, that it's because SKI was invalid.

server/server.go
```diff
@@ -488,8 +494,13 @@ func (w *ecdsaWorker) Do(job interface{}) interface{} {
 	}
 
 	keyLoadBegin := time.Now()
-	if key, ok = w.s.keys.Get(&pkt.Operation); !ok {
-		log.Error(protocol.ErrKeyNotFound)
+	key, err := w.s.keys.Get(&pkt.Operation)
+	if err != nil {
+		log.Errorf("failed to load key: %v", err)
+		w.s.stats.logInternalError(pkt.Opcode)
```
Bren2010: Interesting thought: If a request fails with `logInternalError` or `logInvalid`, then we don't collect metrics on how long the request took to serve. I don't think we do, at least.

cbroglie: That's correct, and that was the intention of the original author: https://github.com/cloudflare/gokeyless/blob/05339602cb41b7859abfd099b9b4dfa4ad19ed9d/server/server.go#L678.

Personally I think it would be valuable to have those timing metrics regardless, with a label for success/failure so we can still slice later if desired. I can make that change now if you agree it would be valuable.

Bren2010: I agree.

server/server.go
```diff
@@ -356,71 +361,60 @@ func (w *otherWorker) Do(job interface{}) interface{} {
...
+			log.Errorf("failed to load key: %v", err)
+			return w.s.makeErrResponse(req, protocol.ErrInternal, requestBegin)
+		} else if key == nil {
+			log.Errorf("failed to load key: %v", protocol.ErrKeyNotFound)
+			return w.s.makeErrResponse(req, protocol.ErrKeyNotFound, requestBegin)
 		}
 		w.s.stats.logKeyLoadDuration(keyLoadBegin)
-		// Reset request beginning time; we don't want to have the request execution
-		// duration include the duration to load the key.
-		requestBegin = time.Now()
```
cbroglie: I disagree, I think we should count this time

server/server.go
```diff
@@ -528,26 +511,23 @@ func (w *ecdsaWorker) Do(job interface{}) interface{} {
...
+		return w.s.makeErrResponse(req, protocol.ErrInternal, requestBegin)
+	} else if key == nil {
+		log.Errorf("failed to load key: %v", protocol.ErrKeyNotFound)
+		return w.s.makeErrResponse(req, protocol.ErrKeyNotFound, requestBegin)
 	}
 	w.s.stats.logKeyLoadDuration(keyLoadBegin)
 
-	// Start measuring the request execution duration here; we don't want to
-	// include the duration to load the key.
-	requestBegin := time.Now()
```
cbroglie: I disagree, I think we should count this time

cbroglie:
@Bren2010 I made the changes we discussed, please review again when you get a chance, thanks

server/metrics.go
```diff
@@ -11,17 +10,14 @@ import (
...
-	connFailures                 prometheus.Counter
-	queuedECDSARequests          prometheus.Gauge
-	queuedOtherRequests          prometheus.Gauge
+	requestExecDuration  *prometheus.HistogramVec
+	requestTotalDuration *prometheus.HistogramVec
+	requests             *prometheus.CounterVec
+	requestsInvalid      *prometheus.CounterVec
+	keyLoadDuration      prometheus.Histogram
+	connFailures         prometheus.Counter
+	queuedECDSARequests  prometheus.Gauge
```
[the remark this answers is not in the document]
cbroglie: Functionally they are equivalent. I probably would use a single metric with labels if I were writing this anew, but there isn't much value in switching now.

server/server.go
```diff
@@ -146,21 +146,21 @@ func DefaultLoadKey(in []byte) (priv crypto.Signer, err error) {
...
+	}
+	priv, found := keys.skis[ski]
+	if found {
+		log.Infof("fetch key with SKI: %s", ski)
+		return priv, nil
 	}
 
-	log.Errorf("couldn't fetch key for %s.", op)
-	return nil, false
+	return nil, fmt.Errorf("keyless: no key with SKI %s", ski)
```
Bren2010: return nil, nil ??

server/metrics.go
```diff
@@ -87,59 +68,37 @@ func newStatistics() *statistics {
...
-			// b) probably a DoS attack to try to get us to generate tons of labels
-			primes = 0
-		}
-		primesStr = fmt.Sprint(primes)
-	}
-	stats.requestExecDurationByOpcode.WithLabelValues(opcode.String(), primesStr).
-		Observe(float64(time.Now().Sub(requestBegin)) / float64(time.Second))
+// including queueing).
+func (stats *statistics) logRequestExecDuration(opcode protocol.Op, requestBegin time.Time, err protocol.Error) {
+	stats.requestExecDuration.WithLabelValues(opcode.String(), err.String()).Observe(time.Since(requestBegin).Seconds())
```
Bren2010: I would just have a "success", 1 for err == nil, 0 for err != nil, or something. err.String() has the potential to create a ton of tags and blow up opentsdb. I also don't like ErrNone.

server/server.go
```diff
@@ -279,20 +279,29 @@ type response struct {
...
 }
 
-func makeErrResponse(req request, err protocol.Error) response {
-	return response{id: req.pkt.ID, op: protocol.MakeErrorOp(err), reqOpcode: req.pkt.Opcode, reqBegin: req.reqBegin}
+func (s *Server) makeErrResponse(req request, err protocol.Error, requestBegin time.Time) response {
+	s.stats.logRequestExecDuration(req.pkt.Opcode, requestBegin, err)
+
+	// This metric can now be derived from the above, and can be removed once
+	// dashboards are updated to use it.
+	s.stats.logInvalid(req.pkt.Opcode)
```
Bren2010: Why don't we just remove it now?

cbroglie: I need to be able to continue using the existing metrics during the rollout of new ones, otherwise I could be blind to new regressions.

server/metrics.go
Bren2010: Ah, I thought `err` was of `type error` but it's of `type protocol.Error`. You're right, it's bounded by the size of the enum, and ErrNone is the type's zero value.

PR #240, opened 2018-07-12
Reduce cardinality of histogram metrics

cbroglie:
The combined cardinality of the labels and buckets was too taxing for
prometheus.

### DIFF
--- a/server/metrics.go
+++ b/server/metrics.go
@@ -21,25 +21,23 @@ type statistics struct {
 }
 
 var (
-	// 1 microsecond as a fraction of 1 second
-	us = 1e-6
-	// buckets starting at 1 microsecond and doubling until reaching a maximum of
-	// ~8 seconds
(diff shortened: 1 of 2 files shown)


server/metrics.go
```diff
@@ -21,25 +21,25 @@ type statistics struct {
...
 
 func newStatistics() *statistics {
 	return &statistics{
 		requestExecDuration: prometheus.NewHistogramVec(prometheus.HistogramOpts{
 			Name:    "keyless_request_exec_duration_per_opcode",
-			Help:    "Time to execute a request not including time in queues, broken down by opcode and error code.",
+			Help:    "Time to execute a request not including time in queues, broken down by type and error code.",
 			Buckets: durationBuckets,
-		}, []string{"opcode", "error"}),
+		}, []string{"type", "error"}),
```
cbroglie: We can still see the breakdown of requests by opcode via the `requests` metric.

PR #242, opened 2018-07-24
Add support for Ed25519

cjpatton:
The opcode takes an arbitrary-length payload. This is because EdDSA signatures can't operate on the hash of the message. These signatures are processed by the otherWorker; since the signature is deterministic, there's no need to precompute random values, as is done with ECDSA.


server/server.go
```diff
@@ -357,6 +359,27 @@ func (w *otherWorker) Do(job interface{}) interface{} {
...
+		if err != nil {
+			log.Errorf("failed to load key with sni=%s ip=%s ski=%v: %v", pkt.Operation.SNI, pkt.Operation.ServerIP, pkt.Operation.SKI, err)
+			return w.s.makeErrResponse(req, protocol.ErrInternal, requestBegin)
+		} else if key == nil {
+			log.Errorf("failed to load key with sni=%s ip=%s ski=%v: %v", pkt.Operation.SNI, pkt.Operation.ServerIP, pkt.Operation.SKI, protocol.ErrKeyNotFound)
+			return w.s.makeErrResponse(req, protocol.ErrKeyNotFound, requestBegin)
+		}
+		w.s.stats.logKeyLoadDuration(keyLoadBegin)
+
+		ed25519Key, ok := key.(ed25519.PrivateKey)
```
Bren2010: You need a contingency for key not being of type ed25519.PrivateKey.

cjpatton:
* We could address the problem of unbounded input length by enforcing a maximum length. For example, we could replace the Ed25519Sign opcode with an opcode for producing TLS 1.3 CertificateVerify messages. The input always has a fixed length that depends on the hash function, so it would be easy to enforce its length.
* I'm not sure I understand what these benchmarks are supposed to say. Are the test inputs the same?

server/server.go
cjpatton: I believe that's handled on line 375. If ok==true, then on line 380, the first argument to ed25519.Sign() is the right type.

cjpatton:
The benchmarks being run look pretty different. I'll put together a proper benchmark so that we're not comparing apples and oranges and report back.

server/server.go
Bren2010: I mean that you need to handle the case where keys.Get() returns a 'keyless' (or some other abstracted) private key. A la line 400 in OpRSADecrypt.
```
if ok {
    // type-specific things
} else {
    // delegate type-specific things to the interface implementation
}
```

cjpatton: Done.

cjpatton:
To be clear, the thing that's being sent over WAN is the transcript hash plus some context. See the example in Sec. 4.4.3 of the TLS spec:
https://tlswg.github.io/tls13-spec/draft-ietf-tls-tls13.html#rfc.section.4.4.3

If you'd like, the client could send ONLY the transcript hash, and the sever could concatenate it to the context string and sign that. (This would save 98 bytes per request.) This would be easy to add just for Ed25519, but to make this general would require some re-engineering of keyless. For one, the client chooses the opcode based on the key type: see `signOpFromSignerOpts()`, which takes as input the `client.PrivateKey` (e.g., as returned by `NewRemoteSigner()`) and a `crypto.SignerOpts`. It could be done, but I'm wary of changing too much.

What do you think? Are you happy with the `OpEd25519Sign` as it is?

server/server.go
```diff
@@ -371,13 +371,22 @@ func (w *otherWorker) Do(job interface{}) interface{} {
 		}
 		w.s.stats.logKeyLoadDuration(keyLoadBegin)
 
-		ed25519Key, ok := key.(ed25519.PrivateKey)
+		if ed25519Key, ok := key.(ed25519.PrivateKey); ok {
+			sig := ed25519.Sign(ed25519Key, pkt.Operation.Payload)
+			return w.s.makeRespondResponse(req, sig, requestBegin)
+		}
+
+		ed25519Key, ok := key.(crypto.Signer)
```
Bren2010: Isn't key already a crypto.Signer?

cjpatton:
For the record: https://gist.github.com/cjpatton/6b72bf72c6134e4b7f882142fbed1b0a

server/server.go
cjpatton: Aww you're right.

cjpatton: Fixed.

PR #244, opened 2018-07-25
Remove keys from DefaultKeystore

cjpatton:
Based on #242 

This adds a function for remove keys from the DefaultKeystore. Keys associated with delegated credentials are short-lived, so a server with a long uptime will need to be able to prune its memory. It's the caller's responsibility to keep track of when keys expired and need to be pruned.


cjpatton:
I rebased this PR, should be Ok to merge now

cjpatton:
We'll revisit this later.

PR #245, opened 2018-07-28
Speed up RPC-over-Keyless

cjpatton:
This adds benchmarks for comparing standard RPC to RPC-over-Keyless. If you checkout https://github.com/cloudflare/gokeyless/commit/255c0f3eb63bc54b7fb6b48798c3aed6da1bbf80 and run `cd tests && go test -bench .`, you'll see that RPC-over-Keyless is up to 10x as slow as RPC:
```
pkg: github.com/cloudflare/gokeyless/tests
BenchmarkKeylessDummyRPC-4   	   10000	    169289 ns/op	   32688 B/op	     471 allocs/op
BenchmarkDummyRPC-4          	  100000	     16991 ns/op	     400 B/op	      12 allocs/op
```
Commit https://github.com/cloudflare/gokeyless/pull/245/commits/6efd7c35b7207a435a7bb2eac6c001ce7a7f999c makes RPC-over-Keyless faster by making encoding and decoding of RPCs stateful. This improves performance by up to a factor of 2:
```
pkg: github.com/cloudflare/gokeyless/tests
BenchmarkKeylessDummyRPC-4   	   20000	     81775 ns/op	   10935 B/op	      48 allocs/op
BenchmarkDummyRPC-4          	  100000	     16408 ns/op	     388 B/op	      12 allocs/op
```
I think there's definitely room for improvement, but it would take some work. This PR is meant to be minimally invasive.

NOTE: this PR changes the wire protocol. If an out-of-date client tries to use RPC with a current server (or vice versa), then the call will fail.

### DIFF
--- a/conn/conn.go
+++ b/conn/conn.go
@@ -307,35 +307,52 @@ func (c *Conn) Ping(data []byte) error {
 // *rpc.Client will cleanup any spawned goroutines, but will not close the
 // underlying connection.
 func (c *Conn) RPC() *rpc.Client {
-	pr, pw := io.Pipe()
-	codec := &clientCodec{c, pr, pw, nil}
-
-	return rpc.NewClientWithCodec(codec)
+	return rpc.NewClientWithCodec(newClientCodec(c))
 }
 
 // clientCodec implements net/rpc.ClientCodec over a connection to a gokeyless
 // server.
 type clientCodec struct {
-	conn *Conn
+	conn     *Conn
+	in       *io.PipeWriter
+	out      bytes.Buffer
+	request  *gob.Encoder
+	response *gob.Decoder
 
-	pr  *io.PipeReader
-	pw  *io.PipeWriter
-	dec *gob.Decoder
+	mtx sync.Mutex
+}
+
+func newClientCodec(c *Conn) *clientCodec {
+	pr, pw := io.Pipe()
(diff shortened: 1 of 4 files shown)


cbroglie:
Changing the wire protocol is unacceptable if it can't be done in a backwards compatible way.

cjpatton:
That's what I figured. One way we could maintain backwards compatibility is to add a new opcode for doing stateful RPC.

But is this really worth doing? Given the current use cases for RPC-over-Keyless, perhaps not. But if this code path is latency sensitive, then to me it seems the current performance is cause for concern.

cjpatton:
For signing DCs using RPC-over-Keyless, a similar benchmark clocks in at 245939 ns/op. Is that Ok, do you think? (Using this speed up gets it down to 155192 ns/op.)

Bren2010:
I think that's great.

cjpatton:
Alrighty :) Go ahead and close this then.

cjpatton:
Actually, I think it's worth at least merging the benchmark: **https://github.com/cloudflare/gokeyless/pull/245/commits/255c0f3eb63bc54b7fb6b48798c3aed6da1bbf80**

cbroglie:
This changes the wire protocol so it cannot be merged. The benchmark included here proves there is no need for optimizations.

PR #246, opened 2018-09-12
Graceful handling of idle sessions being closed

cbroglie:
This also makes the idle timeout configurable, while defaulting to 0 (disabled). It can be configured via the PKCS#11 URI.

Depends on https://github.com/ThalesIgnite/crypto11/pull/10


cbroglie:
Closing in favor of #262

PR #248, opened 2018-10-24
Fix golint import path

golint-fixer:


cbroglie:
Closing in favor of #250.

PR #249, opened 2018-10-24
Fix (and vendor) golint import

cbroglie:


cbroglie:
For some reason github isn't picking up the change from a force push. Closing in favor of #250.

PR #252, opened 2018-11-30
Watson/subcerts

wbl:
This adds support for issuing delegated credentials as well as limiting a connection to only issue delegated credentials.


cmd/gokeyless-cli-tester/gokeyless-cli-tester.go
```diff
@@ -50,6 +60,55 @@ func main() {
...
+		r, err := c.LookupServer(keyserver)
+		if err != nil {
+			log.Fatal("Could not look up keyserver", err)
+		}
+		conn, err := r.Dial(c)
+		if err != nil {
+			log.Fatal("Could not establish connection", err)
+		}
+		rpc := conn.RPC()
+		certfile, err := os.Open(delegateCert)
```
Bren2010: pemblock, err := ioutil.ReadFile(delegateCert) ?

cmd/gokeyless-cli-tester/gokeyless-cli-tester.go
```diff
@@ -50,6 +60,55 @@ func main() {
...
+		}
+
+		ttl := 24 * time.Hour
+		cred, _, err := delegated.NewCredential(uint16(tls.ECDSAWithP256AndSHA256), delegated.VersionTLS13, ttl)
+		req, err := cred.Marshal()
+		if err != nil {
+			log.Fatal("failed to create query", err)
+		}
+		var resp []byte
+		err = rpc.Call("Delegator.Sign", req, &resp) //Richer queries and responses would be a good idea
```
Bren2010: Note 1: "// Richer" (space between)
Note 2: The note is right, you should be able to pass Credential / DelegatedCredential over RPC without marshalling/unmarshalling. Unless you chose not to do that for a reason?

cmd/gokeyless-cli-tester/gokeyless-cli-tester.go
```diff
@@ -50,6 +60,55 @@ func main() {
...
+			log.Fatal("failed in RPC", err)
+		}
+		dc, err := delegated.UnmarshalDelegatedCredential(resp)
+		if err != nil {
+			log.Fatal("failed to parse response")
+		}
+		valid, err := dc.Validate(cert, time.Now())
+		if err != nil {
+			log.Fatal("Error in validation", err)
+		}
```
Bren2010: else if !valid

cmd/gokeyless-cli-tester/gokeyless-cli-tester.go
```diff
@@ -50,6 +60,55 @@ func main() {
...
+		if !valid {
+			log.Fatal("Got invalid response")
+		}
+		log.Fatalf("Got valid response")
+		log.Fatalf("Got response %v", resp)
```
Bren2010: Only the first Fatalf is ever going to come through, and the script is going to exit with status 1. Should be 0 if successful.

fmt.Printf
fmt.Printf
return

cmd/gokeyless-cli-tester/gokeyless-cli-tester.go
```diff
@@ -50,6 +60,55 @@ func main() {
 	if err != nil {
 		log.Fatal("failed to load cert directory:", err)
 	}
+	if delegate {
```
Bren2010: This is basically a standalone function, but you've wrapped it in an if statement. I'd move this code into its own function, and move the c.ScanDir call closer to `for _, priv := range privs` than the invocation of this.

wbl:
I think I've addressed all the comments.

PR #254, opened 2019-03-20
Fixes to make tests independent of the current time

Lekensteyn:
Certificate validation is dependent on the current time, to ensure that tests won't fail in 98 years, make sure that the reference time is fixed.

I've also updated the Travis config while at it, let me know if you prefer a separate PR for that.

The change has additionally been [verified](https://travis-ci.org/Lekensteyn/gokeyless/jobs/509410011) by changing the system time with:
```diff
--- a/tests/trust-check.sh
+++ b/tests/trust-check.sh
@@ -62,6 +62,13 @@ fail() {
     exit 1
 }
 
+# Ensure that tests are run far in the future to ensure that time overrides work
+# as intended.
+if [[ "$(uname -s)" == Linux ]]; then
+    sudo date -s '+200 year'
+    trap "sudo date -s '-200 year'" EXIT
+fi
+
 echo "Checking validation of a valid certificate"
 keyless_connect 2019-01-01T00:00:00Z ||
     fail "Should pass on a valid certificate"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,6 +100,10 @@ test:
 test-nohsm:
 	GODEBUG=cgocheck=2 go test -v -cover -race `go list ./... | grep -v /vendor/`
 
+.PHONY: test-trust
+test-trust: gokeyless
+	tests/trust-check.sh
+
 .PHONY: benchmark-softhsm
 benchmark-softhsm:
 	go test -v -race ./server -bench HSM -args -softhsm2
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -100,6 +100,10 @@ test:
 test-nohsm:
 	GODEBUG=cgocheck=2 go test -v -cover -race `go list ./... | grep -v /vendor/`
 
+.PONY: test-trust
```
Lekensteyn: Ponies... oops, will fix this typo.

PR #258, opened 2019-04-15
Fix race condition between shutdown and new connections

wbl:
Please look this over carefully: there may be subtleties I didn't appreciate when I wrote this.


cbroglie:
Incorporated this in #259

PR #259, opened 2019-04-16
Add concurrent tests and fix races they uncover

cbroglie:
The integration tests were refactored to use dedicated server instances per test so we can verify the server shuts down cleanly after each test. And we want to exercise the connection management code as much as possible, so we also disable connection multiplexing by default during these tests.


cbroglie:
Oh great, the locally passing test fails on Travis...

server/server.go
```diff
@@ -682,69 +687,37 @@ func (s *Server) Serve(l net.Listener) error {
...
-			if s.config.isLimited(tconn) {
-				log.Debug("Connection is limited")
-				conn = newConn(s, c.RemoteAddr().String(), tconn, timeout, s.wp.Limited, s.wp.Limited)
-			} else {
-				conn = newConn(s, c.RemoteAddr().String(), tconn, timeout, s.wp.ECDSA, s.wp.Other)
-			}
+	s.mtx.Lock()
+	if s.shutdown {
+		s.mtx.Unlock()
+		log.Debugf("connection %v: rejected b/c server is shutting down", c.RemoteAddr())
```
wbl: Do we ever close it?

cbroglie: Nope, good point

server/internal/client/client.go
```diff
@@ -185,10 +186,9 @@ type ConnHandle struct {
 // the other waits of the results of these jobs and writes them to the client.
 func SpawnConn(conn Conn) *ConnHandle {
 	c := &ConnHandle{
-		conn: conn,
-		// responses needs to have an extra slot for the sentinal value sent by
-		// Destroy
-		responses: make(chan interface{}, maxOutstandingRequests+2),
+		conn:      conn,
+		responses: make(chan interface{}, maxOutstandingRequests+1),
```
wbl: Do we still need the extra slot?

server/internal/client/client.go
```diff
@@ -228,18 +228,18 @@ func SpawnConn(conn Conn) *ConnHandle {
...
-		// is undefined - but it's nice and makes it less likely that bugs will slip
-		// by unnoticed.
-		panic("destroy already-destroyed ClientHandle")
-	}
 	c.destroy()
 	c.wg.Wait()
 }
 
 func (c *ConnHandle) destroy() {
+	// This can be called explicitly, b/c the health check has failed, or both.
```
wbl: This comment broke my English parser temporarily where I read a parenthetical phrase not a list.  Maybe "This function could be called both by the health checker and explicitly" would flow a bit better?

server/server.go
```diff
@@ -682,58 +675,55 @@ func (s *Server) Serve(l net.Listener) error {
...
+	s.listeners[l][handle] = struct{}{}
+	s.mtx.Unlock()
+	log.Debugf("%s: spawned", connStr)
 
-			mapMtx.Lock()
-			conns[handle] = true
-			mapMtx.Unlock()
+	// Block here until the connection and associated goroutines have completed.
+	handle.Wait()
+	log.Debugf("%s: removed", connStr)
```
wbl: Hang on, we're logging this line before we actually removed the connection. Maybe we should reverse the order to help in debugging.

server/server.go
```diff
@@ -791,19 +781,30 @@ func (s *Server) UnixListenAndServe(path string) error {
...
+
+	s.shutdown = true
+	for l, conns := range s.listeners {
+		delete(s.listeners, l)
+
+		log.Debugf("Shutting down %v; closing %d active connections", l.Addr().String(), len(conns))
+		for conn := range conns {
+			conn.Destroy()
+		}
 		l.Close()
```
wbl: Maybe we should close the listener socket ahead of sweeping through the connections. As is while the cleanup is happening lots of handshakes can still be happening that are sort of useless since they will be rejected immediately.

server/internal/client/client.go
cbroglie: I changed it from +2 to +1 b/c I wasn't sure why it was +2 in the first place (rather than the +1 the comment explained). But now that I think about it may be because both Destroy and the health check goroutine could call destroy? In which case I can remove the +1 here.

server/internal/client/client.go
cbroglie: haha sure

server/server.go
cbroglie: I thought about that, but the advantage of doing it this way is it's clear in the logs why these connections are being closed. And in practice the way we do graceful reloads the new process should already be handling the new connections before we get here.

server/server.go
cbroglie: Are you sure? Once `Wait` returns it means both the reader and writer goroutines have exited.

wbl: We killed it, but we didn't take it out of the structure yet.

server/server.go
wbl: Why would calling l.Close() before the loop change the logging? Not a big deal, I'm just a bit confused.

cbroglie: B/c of LogConnErr:
```go
func (c *conn) LogConnErr(err error) {
	if err == nil { // We're destroying the connection
		log.Debugf("connection %v: server closing connection %s", c.name, c.stats)
	} else if err == io.EOF {
		log.Debugf("connection %v: closed by client %s", c.name, c.stats)
	} else {
		c.s.stats.logConnFailure()
		log.Errorf("connection %v: encountered error: %v %s", c.name, err, c.stats)
	}
}
```
It's the difference between logging "server closing connection" as a debug message and "encountered error" as an error message.

server/server.go
cbroglie: Oh that's what you mean by removed. I generally care more about when the connection is closed rather than actually removed from the map, but I can see value in either. I'll log both steps.

server/internal/client/client.go
wbl: That was it

server/server.go
cbroglie: I wrote a small program to test this and I'm wrong, closing the listener does not close the connections. Will fix.

PR #260, opened 2019-05-10
Add Custom Function OpCode

lgarofalo:
This change adds a new operation to the keyless server that allows a custom function to be defined and applied to process that operation.

The interface allows the custom function to accept the keyless operation payload bytes and return bytes for the response payload.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ The following tag values are possible for items:
     0x03 - Client's IP address,
     0x11 - Opcode,
     0x12 - Payload,
+    0x13 - CustomFuncName, (for use with opcode 0x24)
 
 A requests contains a header and the following items:
 
@@ -84,6 +85,7 @@ The following opcodes are supported in the opcode item:
     0x15 - operation: ECDSA sign SHA256
     0x16 - operation: ECDSA sign SHA384
     0x17 - operation: ECDSA sign SHA512
+    0x23 - operation: RPC
     0x24 - operation: Custom Function
     0x35 - operation: RSASSA-PSS sign SHA256
     0x36 - operation: RSASSA-PSS sign SHA384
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -84,6 +84,7 @@ The following opcodes are supported in the opcode item:
     0x15 - operation: ECDSA sign SHA256
     0x16 - operation: ECDSA sign SHA384
     0x17 - operation: ECDSA sign SHA512
+    0x24 - operation: Custom Function
```
wbl: OpRPC isn't in this README. Should we add it?

protocol/protocol.go
```diff
@@ -151,6 +155,8 @@ const (
 	ErrCertNotFound
 	// ErrExpired indicates that the sealed blob is no longer unsealable.
 	ErrExpired
+	// ErrCustomOpUndefined indicates that the custom opcode was invoked but no definition was provided to the server configuration.
```
cbroglie: I think it's ok to reuse `ErrBadOpcode` here, and would like to avoid growing this list unless strictly required.

protocol/protocol.go
```diff
@@ -95,6 +95,8 @@ const (
 	OpUnseal Op = 0x22
 	// OpRPC executes an arbitrary exported function on the server.
 	OpRPC Op = 0x23
+	// OpCustom requests a custom operation that can be defined by a function set in the server configuration
+	OpCustom Op = 0x24
```
cbroglie: This doesn't seem very extensible to me. I'd prefer to see this named something like `OpCustom1` so it's clear this has a single purpose, or to add another tag to identify what kind of custom operation is being performed.

lgarofalo: Agree, I will go the route of adding a function name to implementation mapping to the configuration of the opcode, and expect that function name in the payload.

protocol/protocol.go
lgarofalo: That sounds good.

README.md
lgarofalo: Will do.

PR #261, opened 2019-05-16
Add support for an additional payload field

cbroglie:
Also re-ran [stringer](https://godoc.org/golang.org/x/tools/cmd/stringer) to regenerate protocol_string.go after the addition of `OpCustom`.


nickysemenza:
:shipit:

PR #262, opened 2019-05-19
Upgrade to v1.2.0 version of crypto11 library

cbroglie:
The V1 release was a a major rewrite: https://github.com/ThalesIgnite/crypto11/pull/40.


dmjones:
I corrected the erroneous tag value `v.1.0.0` by deleting it and renaming it to `v1.0.0`. On reflection, that may not have been great Git etiquette and presumably breaks this commit.

I can't find decent information on how Go module versioning would handle a ".1.0.0" release (in terms of ordering), so I was reluctant to leave the tag in place, in case it caused us loads of version issues in the future. Or in case Go quietly changes its behaviour in this regard, breaking even more builds.

Apologies for the mistake. Happy to take a different approach to fixing it if you have suggestions. (Which we should discuss back at https://github.com/ThalesIgnite/crypto11/pull/40.

cbroglie:
@dmjones Thanks for fixing, deleting the old tag is no problem for me

.travis.yml
```diff
@@ -16,15 +18,3 @@ go_import_path: github.com/cloudflare/gokeyless
 
 script:
   - make vet lint test-nohsm test-trust
-
-notifications:
```
cbroglie: I don't think email notifications are providing value

PR #272, opened 2020-06-24
update vulnerable DNS dependency

wbl:


cbroglie:
Could use a `go mod tidy`

PR #275, opened 2020-07-07
add certificate expiration metrics for client certificates used in handshake

mitalirawat:
We want to expose certification expiration metrics of any certificate in memory in gokeyless. This PR will create a package for defining this metric and use it to record expiration times of client certificates sent from any service that connects to gokeyless.
The package can be imported and reused for certificates in any repo. These metrics will be sent to promoetheus where alerts can be configured based on them.

### DIFF
--- a/certmetrics/metrics.go
+++ b/certmetrics/metrics.go
@@ -1,0 +1,51 @@
+// Package certmetrics will be used to register and emit metrics for certificates in memory
+package certmetrics
+
+import (
+	"crypto/x509"
+	"sort"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var certificateExpirationTimes = promauto.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Name: "certificate_expiration_timestamp_seconds",
+		Help: "Expiration times of gokeyless certs",
+	},
+	[]string{"serial_no", "cn", "hostnames", "ca", "server", "client"},
+)
+
+// Observe takes in a list of certs and emits its expiration times
+func Observe(certs ...*x509.Certificate) {
+	for _, cert := range certs {
+		hostnames := cert.DNSNames
+		sort.Strings(hostnames)
+		labels := prometheus.Labels{
+			"serial_no": cert.SerialNumber.String(),
+			"cn":        cert.Subject.CommonName,
+			"hostnames": strings.Join(hostnames, ","),
+			"ca":        boolToBinaryString(cert.IsCA),
+			"server":    containsKeyUsage(cert.ExtKeyUsage, x509.ExtKeyUsageServerAuth),
+			"client":    containsKeyUsage(cert.ExtKeyUsage, x509.ExtKeyUsageClientAuth)}
+		certificateExpirationTimes.With(labels).Set(float64(cert.NotAfter.Unix()))
+	}
+}
+
+func boolToBinaryString(val bool) string {
+	if val {
+		return "1"
+	}
+	return "0"
+}
+
+func containsKeyUsage(a []x509.ExtKeyUsage, x x509.ExtKeyUsage) string {
+	for _, e := range a {
+		if e == x || e == x509.ExtKeyUsageAny {
+			return "1"
+		}
+	}
+	return "0"
+}
--- a/certmetrics/metrics.go
+++ b/certmetrics/metrics.go
@@ -21,17 +21,20 @@ var certificateExpirationTimes = promauto.NewGaugeVec(
 // Observe takes in a list of certs and emits its expiration times
 func Observe(certs ...*x509.Certificate) {
 	for _, cert := range certs {
-		hostnames := cert.DNSNames
-		sort.Strings(hostnames)
-		labels := prometheus.Labels{
-			"serial_no": cert.SerialNumber.String(),
-			"cn":        cert.Subject.CommonName,
-			"hostnames": strings.Join(hostnames, ","),
-			"ca":        boolToBinaryString(cert.IsCA),
-			"server":    containsKeyUsage(cert.ExtKeyUsage, x509.ExtKeyUsageServerAuth),
-			"client":    containsKeyUsage(cert.ExtKeyUsage, x509.ExtKeyUsageClientAuth)}
-		certificateExpirationTimes.With(labels).Set(float64(cert.NotAfter.Unix()))
+		certificateExpirationTimes.With(getPrometheusLabels(cert)).Set(float64(cert.NotAfter.Unix()))
 	}
+}
+
+func getPrometheusLabels(cert *x509.Certificate) prometheus.Labels {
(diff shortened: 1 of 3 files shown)


certmetrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+	for _, cert := range certs {
+		hostnames := append([]string(nil), cert.DNSNames...)
+		sort.Strings(hostnames)
+		labels := prometheus.Labels{
+			"serial_no": cert.SerialNumber.String(),
+			"cn":        cert.Subject.CommonName,
+			"hostnames": strings.Join(hostnames, ","),
+			"ca":        boolToBinaryString(cert.IsCA),
+			"server":    searchKeyUsage(cert.ExtKeyUsage, x509.ExtKeyUsageServerAuth),
+			"client":    searchKeyUsage(cert.ExtKeyUsage, x509.ExtKeyUsageClientAuth)}
```
mitalirawat, line 32: can create a map and use it for search by index if that sounds better

server/server.go
```diff
@@ -644,7 +646,9 @@ func (s *Server) spawn(l net.Listener, c net.Conn) {
 		tconn.Close()
 		return
 	}
-	limited, err := s.config.isLimited(tconn.ConnectionState())
+	connState := tconn.ConnectionState()
+	certmetrics.Observe(connState.PeerCertificates...)
```
mitalirawat, line 650: not sure if I also need to observe certificates in Verified Chains of ConnectionState here

certmetrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func boolToBinaryString(val bool) string {
+	if val {
+		return "1"
+	}
+	return "0"
+}
+
+func searchKeyUsage(a []x509.ExtKeyUsage, x x509.ExtKeyUsage) string {
```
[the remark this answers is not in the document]
mitalirawat: Yea I thought of `boolToBinaryString(searchKeyUsage(cert.ExtKeyUsage, x509.ExtKeyUsageServerAuth))` but felt this does an extra step of processing for converting bool/int to string. I am thinking if we need to add more code that could reuse this function, we could modify this then.

certmetrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+		Name: "certificate_expiration_timestamp_seconds",
+		Help: " Expiration times of gokeyless certs",
+	},
+	[]string{"serial_no", "cn", "hostnames", "ca", "server", "client"},
+)
+
+// Observe takes in a list of certs and emits its expiration times
+func Observe(certs ...*x509.Certificate) {
+	for _, cert := range certs {
+		hostnames := append([]string(nil), cert.DNSNames...)
```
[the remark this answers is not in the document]
mitalirawat: not much of a difference except that the expr will give me a nil slice instead of an empty slice if DNSNames is empty.

certmetrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func boolToBinaryString(val bool) string {
+	if val {
+		return "1"
+	}
+	return "0"
+}
+
+func containsKeyUsage(a []x509.ExtKeyUsage, x x509.ExtKeyUsage) string {
```
cbroglie, line 44: nit: I would make this return a `bool`, and then wrap the result in a call to `boolToBinaryString`, ex.:
```go
"server":    boolToBinaryString(containsKeyUsage(cert.ExtKeyUsage, x509.ExtKeyUsageServerAuth)),
```

server/server.go
cbroglie: It's good to include the full chain here 👍

certmetrics/metrics.go
cbroglie: oops @nickysemenza made this same comment

server/server.go
cbroglie: My apologies, I misunderstood your comment. I thought you were asking whether we needed the full chain from `PeerCertificates` or just the leaf cert (i.e. `PeerCertificates...` vs. `PeerCertificates[0]`), and my comment was supporting the use of `PeerCertificates...`.

I think in practice `PeerCertificates` will be the same as `VerifiedChains` (but I think verified chains will also include the root), so I think it would be best to just observe `PeerCertificates...` here. Worth checking with @maxnystrom.

mitalirawat: makes sense, ideally `PeerCertificates` chain should be one of the chains in `VerifiedChains` so should just observe certs sent by peer. Also, wondering why root wouldnt be a part of `PeerCertificates` or why we should not include root for observation

cbroglie: > Also, wondering why root wouldnt be a part of PeerCertificates 

When a TLS handshake is initiated, the server sends down its leaf certificate, along with any intermediates. The client then checks if it can build a chain to that leaf certificate given the intermediates and the roots it already has embedded in its trust store. The server doesn't include the CA certificate b/c the client must already have it for the handshake to succeed.

With client auth, it's basically the reverse. The server must already have the root CA certificate(s) required to build a chain of trust to the client leaf certificate, or the handshake will fail. Hence I expect `PeerCertificates` to not include the root (though I think it's valid to include it, it would just be ignored by the server).

> or why we should not include root for observation

I didn't mean to imply it would be a problem to include it here, I was just stating my understanding of how these variables differ. Including it won't be a problem, though we know from above it should have already been observed when the server configured the root CAs it allows for client auth.

`VerifiedChains` may also only contain a subset of the certificates from `PeerCertificates`, if the client is sending intermediates the server doesn't need to build the chain of trust. You could argue we only need to report on the relevant certificates, but I think it's best to just report everything the client is sending in `PeerCertificates`.

certmetrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+		Help: "Expiration times of gokeyless certs",
+	},
+	[]string{"serial_no", "cn", "hostnames", "ca", "server", "client"},
+)
+
+// Observe takes in a list of certs and emits its expiration times
+func Observe(certs ...*x509.Certificate) {
+	for _, cert := range certs {
+		hostnames := cert.DNSNames
+		sort.Strings(hostnames)
```
cbroglie, line 25: This is sorting `cert.DNSNames` in place, it's not a copy: https://goplay.space/#EOmjxbq7se2.

This could cause data races depending on how this function is used. I think it's best to skip the sort here, as the order is already deterministic based on the cert contents.

server/server.go
maxnystrom: @cbroglie is correct, though https://tools.ietf.org/html/rfc5246#section-7.4.6 indicates that the client certificate is singular, so the client must only send the leaf (end-entity). 

Though, like server certificates, a certificate chain is only valid if all the certificates in the chain are valid. And I have low confidence that all clients only send the leaf so checking `VerifiedChains` makes sense.

certmetrics/metrics.go
mitalirawat: idk how i missed it, i expected this to be references and then tested to ensure. But it turned out to be a copy so kept this, clearly i was sleeping.

PR #276, opened 2020-09-09
allow Unmarshaling of protocol.Operation with unknown tags

nickysemenza:


wbl (approved):

protocol/protocol.go
```diff
@@ -642,8 +642,6 @@ func (o *Operation) UnmarshalBinary(body []byte) error {
 			// ignore padding
 		case TagCustomFuncName:
 			o.CustomFuncName = string(data)
-		default:
-			return fmt.Errorf("unknown tag: %02x", tag)
```
cbroglie, line 646: I think you want to continue here instead. See the comment on the next line.

cbroglie (changes requested):

protocol/protocol.go
nickysemenza: Ah I think you're right, I was looking at the comment for `TagPadding` but I guess that's different because it's both known and ignored.

cbroglie (approved):

protocol/protocol.go
```diff
@@ -643,7 +643,8 @@ func (o *Operation) UnmarshalBinary(body []byte) error {
 		case TagCustomFuncName:
 			o.CustomFuncName = string(data)
 		default:
-			return fmt.Errorf("unknown tag: %02x", tag)
+			// Silently ignore any unkown tags (to allow for new tags to be gradually added to the protocol).
```
cbroglie, line 646: typo: unkown

PR #279, opened 2020-09-20
Fix ZoneID json tag

bvwells:
Fix ZoneID json tag


bvwells:
Thanks @nickysemenza for the review and merge!

PR #281, opened 2021-01-20
replace Travis with Actions

nickysemenza:
Travis free plan was reduced significantly (builds are queued for hours), GH Actions for public repos is much better.


wbl (approved):
Yup lots of errors to clean up, which means its working

mihirjham (approved):

lgarofalo (approved):

PR #285, opened 2021-03-05
feat: support using private keys on Azure Key Vault / Managed HSM

nickysemenza:


cmd/gokeyless/gokeyless.go
```diff
@@ -216,7 +216,7 @@ func main() {
 		fmt.Println("gokeyless version", version)
 		os.Exit(0)
 	case manualMode && configMode:
-		log.Fatal("can't specify both -manual-activation and -config-only!")
+		log.Fatal("can't specify both --manual_activation and --config_only!")
```
cbroglie, line 219: It actually accepts both forms ftr

nickysemenza: Oh, I fixed too much I guess. I was trying to solve `unknown shorthand flag: 'm' in -manual-activation`, but turns you need to do `--manual-activation` (or `--manual_activation`), `config-only` accidentally works with just one dash because it has `-c` registered as a shorthand. tl;dr I just need to change `-` to `--` in the error messages.

cbroglie: 👍

server/server.go
```diff
@@ -137,7 +142,7 @@ func (keys *DefaultKeystore) Add(op *protocol.Operation, priv crypto.Signer) err
 
 	keys.skis[ski] = priv
 
-	log.Debugf("add signer with SKI: %v", ski)
+	log.Debugf("add signer with SKI: %v (https://crt.sh/?ski=%v)", ski, ski)
```
wbl, line 145: Neat idea that makes life just that little bit better!

server/server.go
```diff
@@ -113,14 +114,18 @@ func (keys *DefaultKeystore) AddFromFile(path string, LoadKey func([]byte) (cryp
 // AddFromURI loads all keys matching the given PKCS#11 URI to the keystore. LoadURI
```
wbl, line 114: This comment needs updating.

server/server.go
```diff
@@ -113,14 +114,18 @@ func (keys *DefaultKeystore) AddFromFile(path string, LoadKey func([]byte) (cryp
...
 // is called to parse the URL, connect to the module, and populate a crypto.Signer,
 // which is stored in the Keystore.
-func (keys *DefaultKeystore) AddFromURI(uri string, LoadURI func(string) (crypto.Signer, error)) error {
+func (keys *DefaultKeystore) AddFromURI(uri string, LoadPKCS11URI func(string) (crypto.Signer, error)) error {
 	log.Infof("loading %s...", uri)
-
-	priv, err := LoadURI(uri)
+	var priv crypto.Signer
+	var err error
+	if azure.IsKeyVaultURI(uri) {
```
wbl, line 121: PKCS11 has a scheme registered in RFC 7512 https://tools.ietf.org/html/rfc7512#page-4. Now that we're generalizing this, I think it makes sense to check for that scheme ahead of calling the LoadPKCS11URI, for better uniformity of errors, and it makes more sense once we add a third option (which by the one, none, or many rule, we will).

wbl (approved):
A few minor things, otherwise LGTM

server/server.go
nickysemenza: oh awesome, I wasn't sure if the docs indicating `pkcs:` was an accident or not :)

nickysemenza: hah turns out there's already a regex in the code for it, I moved it so it's not behind the buildflags in 84aadce03bc68cd82e4aac686cb6764f5f6877dd

mihirjham (approved):

PR #286, opened 2021-03-13
feat: support for using keys in google KMS

nickysemenza:
Very similar to #285, but with google this time.


mihirjham (approved):

internal/azure/azure.go
```diff
@@ -90,15 +90,15 @@ func (k KeyVaultSigner) Sign(_ io.Reader, digest []byte, opts crypto.SignerOpts)
 
 	signed, err := k.client.Sign(context.Background(), k.baseURL, k.keyName, k.keyVersion, keyvault.KeySignParameters{Algorithm: algo, Value: &payload})
 	if err != nil {
-		return nil, fmt.Errorf("failed to sign against azure: %w", err)
+		return nil, fmt.Errorf("azure: failed to sign against azure: %w", err)
```
wbl, line 93: Azure now appears twice, we might want to rewrite this message

internal/azure/azure.go
```diff
@@ -90,15 +90,15 @@ func (k KeyVaultSigner) Sign(_ io.Reader, digest []byte, opts crypto.SignerOpts)
...
 	}
 
 	res, err := base64.RawURLEncoding.DecodeString(*signed.Result)
 	if err != nil {
 		return nil, fmt.Errorf("failed to base64 recode result: `%s` %w", *signed.Result, err)
 	}
 
-	log.Infof("signed %d bytes with %s for %s key %s", len(digest), algo, k.keyType, k.keyName)
+	log.Infof("azure: signed %d bytes with %s for %s key %s", len(digest), algo, k.keyType, k.keyName)
```
wbl, line 101: Info or debug? I think debug, but if there are reasons for Info that's fine.

internal/google/google.go
```diff
@@ -0,0 +1,149 @@
...
+	log.Infof("google: signed %d bytes with %s for %s key %s", len(digest), opts.HashFunc().String(), k.keyType, k.name)
+
+	return result.Signature, nil
+
+}
+
+// IsKMSResource attempts to identify if a name is a KMS `Key version`
+// format specified at https://cloud.google.com/kms/docs/resource-hierarchy#retrieve_resource_id
+func IsKMSResource(name string) bool {
+	return strings.Contains(name, "/keyRings/") && strings.Contains(name, "/cryptoKeyVersions/") && len(strings.Split(name, "/")) == 10
```
wbl, line 110: This seems nastily conflictable with other things. Don't think there is much we can do about it, other than make sure this is a best effort "hey, you seem to have done something funny" error.

wbl (approved):
Nothing major, but we really should send in a request.

README.md
```diff
@@ -159,12 +161,22 @@ Note you must provide exactly one of the `token`, `serial`, or `slot-id` attribu
...
 ```
 If gokeyless is running in a VM with Managed Services enabled, auth works out of the box. Otherwise, credentials can also be specified with an env var containing the path to a file. (env vars are defined [here](https://pkg.go.dev/github.com/Azure/go-autorest/autorest/azure/auth#pkg-constants))
 The required roles are `/keys/read/action` and `/keys/sign/action`
 
+### Google Cloud KMS
+
+Private keys can also be stored in Google Cloud's [key management offerings](https://cloud.google.com/security-key-management)
+```
+private_key_stores:
+    - uri: projects/abc/locations/us-west1/keyRings/xyz/cryptoKeys/example-key/cryptoKeyVersions/3
```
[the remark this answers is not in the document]
nickysemenza: so they actually do have a full URI format, but it's _schemaless_ https://cloud.google.com/apis/design/resource_names#full_resource_name. That being said, it's not supported for the KMS API (`name does not match pattern: /^projects\/[^/]+\/locations\/[^/]+\/keyRings\/[^/]+\/cryptoKeys\/[^/]+\/cryptoKeyVersions\/[^/]+$/`), only for these: https://cloud.google.com/iam/docs/full-resource-names and for IAM usage. That being said, I think this is fine as-is since we can strengthen the `IsKMS` resource check if we ever add any additional providers that get even close to conflicting, which is probably unlikely since everyone names everything slightly differently.

internal/azure/azure.go
nickysemenza: I think all of them should probably Debug or Error

README.md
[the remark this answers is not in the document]
wbl: I feel sad now

maxnystrom (approved):

PR #290, opened 2021-06-11
Separate forwarding and expand concurrency

wbl:
Experience has shown that certain applications require distinctions
among their worker pools and a high concurrency per connection.

### DIFF
--- a/server/internal/client/client.go
+++ b/server/internal/client/client.go
@@ -110,7 +110,7 @@ import (
 // by the user, and once by this monitor goroutine.
 
 const (
-	maxOutstandingRequests = 1024
+	maxOutstandingRequests = 16384
 )
 
 // A note on using Destroy:
--- a/server/internal/client/client.go
+++ b/server/internal/client/client.go
@@ -109,10 +109,6 @@ import (
 // must be safe for the underlying destroy procedure to be invoked twice - once
 // by the user, and once by this monitor goroutine.
 
-const (
-	maxOutstandingRequests = 16384
-)
-
 // A note on using Destroy:
 //
 // We need some mechanism for the code to decide that it wants to kill an
@@ -184,12 +180,12 @@ type ConnHandle struct {
 // SpawnConn spawns a pair of goroutines to handle requests from the client. One
 // goroutine reads jobs from the client and submits them a worker pool, while
 // the other waits of the results of these jobs and writes them to the client.
-func SpawnConn(conn Conn) *ConnHandle {
+func SpawnConn(conn Conn, maxOutstanding int) *ConnHandle {
 	c := &ConnHandle{
 		conn:      conn,
-		responses: make(chan interface{}, maxOutstandingRequests),
+		responses: make(chan interface{}, maxOutstanding),
 		done:      make(chan struct{}),
-		blocker:   newBlocker(maxOutstandingRequests),
+		blocker:   newBlocker(maxOutstanding),
 	}
 	c.wg.Add(2)
 
--- a/server/server.go
+++ b/server/server.go
@@ -691,7 +691,7 @@ func (s *Server) spawn(l net.Listener, c net.Conn) {
 		tconn.Close()
 		return
 	}
-	handle := client.SpawnConn(conn)
+	handle := client.SpawnConn(conn, s.config.maxConnPendingRequests)
 	s.listeners[l][handle] = struct{}{}
 	s.mtx.Unlock()
 	log.Debugf("%s: spawned", connStr)
@@ -798,6 +798,7 @@ type ServeConfig struct {
 	otherWorkers            int
 	remoteWorkers           int
 	limitedWorkers          int
+	maxConnPendingRequests  int
 	tcpTimeout, unixTimeout time.Duration
 	isLimited               func(state tls.ConnectionState) (bool, error)
 	customOpFunc            CustomOpFunction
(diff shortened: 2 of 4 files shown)


server/worker_pool.go
```diff
@@ -46,13 +48,17 @@ func newWorkerPool(s *Server) (*workerPool, error) {
...
+	var remotes []worker.Worker
 	var limiteds []worker.Worker
 	for i := 0; i < s.config.rsaWorkers; i++ {
 		rsas = append(rsas, newKeylessWorker(s, fmt.Sprintf("rsa-%v", i)))
 	}
 	for i := 0; i < s.config.ecdsaWorkers; i++ {
 		ecdsas = append(ecdsas, newKeylessWorker(s, fmt.Sprintf("ecdsa-%v", i)))
 	}
+	for i := 0; i < s.config.remoteWorkers; i++ {
+		remotes = append(remotes, newKeylessWorker(s, fmt.Sprintf("other-%v", i)))
```
tanyav2, line 60: This should be `remote-%v`

server/worker_pool.go
```diff
@@ -80,6 +87,7 @@ func newWorkerPool(s *Server) (*workerPool, error) {
 				serverUtilization.WithLabelValues("rsa").Set(float64(wp.RSA.Busy()) / float64(s.config.rsaWorkers))
 				serverUtilization.WithLabelValues("ecdsa").Set(float64(wp.ECDSA.Busy()) / float64(s.config.ecdsaWorkers))
 				serverUtilization.WithLabelValues("other").Set(float64(wp.Other.Busy()) / float64(s.config.otherWorkers))
+				serverUtilization.WithLabelValues("remote").Set(float64(wp.Remote.Busy()) / float64(s.config.remoteWorkers))
```
tanyav2, line 90: I can't add a comment on Line 78, but `for _, label := range []string{"rsa", "ecdsa", "other", "limited"} ` should also have `remote`

tanyav2: Same with line 100, `wp.Remote.Destroy()` might be necessary, though we somehow don't destroy the RSA pool?

server/worker_pool.go
```diff
@@ -46,13 +48,17 @@ func newWorkerPool(s *Server) (*workerPool, error) {
 	var ecdsas []worker.Worker
 	var rsas []worker.Worker
 	var others []worker.Worker
+	var remotes []worker.Worker
```
tanyav2, line 51: Missing error check for remote workers on line 40: 
```
	if s.config.rsaWorkers <= 0 ||
		s.config.ecdsaWorkers <= 0 ||
		s.config.otherWorkers <= 0 {
		return nil, fmt.Errorf("non-zero number of RSA, ECDSA, and Other workers is required")
	}
```

tanyav2:
Line 625 in `server.go` is missing pool selection for remote:
```
func (s *poolSelector) SelectPool(pkt *protocol.Packet) *worker.Pool {
	if s.limited {
		return s.wp.Limited
	}
	switch s.wp.selector(pkt) {
	case PoolRSA:
		return s.wp.RSA
	case PoolECDSA:
		return s.wp.ECDSA
	default:
		return s.wp.Other
	}
}
```

tanyav2:
Line 834 of `server.go`, the default is other, this should be change to disambiguate between geokdl ops and other ops: 
```
func defaultPoolSelector(pkt *protocol.Packet) WorkerPoolType {
	switch pkt.Operation.Opcode {
	case protocol.OpRSADecrypt, protocol.OpRSASignMD5SHA1,
		protocol.OpRSASignSHA1, protocol.OpRSASignSHA224,
		protocol.OpRSASignSHA256, protocol.OpRSASignSHA384,
		protocol.OpRSASignSHA512, protocol.OpRSAPSSSignSHA256,
		protocol.OpRSAPSSSignSHA384, protocol.OpRSAPSSSignSHA512:
		return PoolRSA
	case protocol.OpECDSASignMD5SHA1, protocol.OpECDSASignSHA1,
		protocol.OpECDSASignSHA224, protocol.OpECDSASignSHA256,
		protocol.OpECDSASignSHA384, protocol.OpECDSASignSHA512:
		return PoolECDSA
	default:
		return PoolOther
	}
}
```

wbl:
Yeah there are a whole bunch of places I missed where I need to add this pool as well...

server/internal/client/client.go
```diff
@@ -110,7 +110,7 @@ import (
 // by the user, and once by this monitor goroutine.
 
 const (
-	maxOutstandingRequests = 1024
+	maxOutstandingRequests = 16384
```
cbroglie, line 113: In addition to increasing the default, it's probably a good idea to make this configurable at runtime,

wbl:
Note that we don't use the default pool selector internally.

tanyav2 (approved):

server/server.go
```diff
@@ -820,14 +824,16 @@ func DefaultServeConfig() *ServeConfig {
...
-		isLimited:      func(state tls.ConnectionState) (bool, error) { return false, nil },
-		poolSelector:   defaultPoolSelector,
+		rsaWorkers:             n,
+		ecdsaWorkers:           n,
+		otherWorkers:           2,
+		remoteWorkers:          2,
+		limitedWorkers:         0,
+		tcpTimeout:             defaultTCPTimeout,
+		unixTimeout:            defaultUnixTimeout,
+		maxConnPendingRequests: 1024,
```
tanyav2, line 834: We were supposed to increase this to 16684

tanyav2 (changes requested):

tanyav2 (approved):

server/worker_pool.go
```diff
@@ -39,20 +41,25 @@ func newWorkerPool(s *Server) (*workerPool, error) {
...
 	var limiteds []worker.Worker
 	for i := 0; i < s.config.rsaWorkers; i++ {
 		rsas = append(rsas, newKeylessWorker(s, fmt.Sprintf("rsa-%v", i)))
 	}
 	for i := 0; i < s.config.ecdsaWorkers; i++ {
 		ecdsas = append(ecdsas, newKeylessWorker(s, fmt.Sprintf("ecdsa-%v", i)))
 	}
+	for i := 0; i < s.config.remoteWorkers; i++ {
+		remotes = append(remotes, newKeylessWorker(s, fmt.Sprintf("remotes-%v", i)))
```
tanyav2, line 61: nit: `remote` instead of `remotes`

mihirjham (approved):

server/server.go
cbroglie: This change maintains the existing default while making it configurable

cbroglie (approved):

server/worker_pool.go
```diff
@@ -99,6 +108,9 @@ func (wp *workerPool) Destroy() {
 	// Destroy the pools
 	wp.Other.Destroy()
 	wp.ECDSA.Destroy()
+	wp.RSA.Destroy()
```
armfazh, line 111: hmm, rsa pool wasn't destroyed before.

wbl: nothing bad happened from it AFAIK,  but who knows?

PR #292, opened 2021-06-21
Watson/revamp conns

wbl:
This restructures the way we execute requests: instead of using a pool of worker routines, we spawn a goroutine for each request, and use a per-connection mutex to serialize the responses.


server/server.go
```diff
@@ -691,19 +511,43 @@ func (s *Server) spawn(l net.Listener, c net.Conn) {
...
-	handle := client.SpawnConn(conn, s.config.maxConnPendingRequests)
-	s.listeners[l][handle] = struct{}{}
+
+	s.listeners[l][tconn] = struct{}{}
 	s.mtx.Unlock()
-	log.Debugf("%s: spawned", connStr)
+	log.Debugf("%s: serving", connStr)
+	handler := &handler{
+		name:        connStr,
+		outstanding: s.config.maxConnPendingRequests,
```
tanyav2, line 520: Shouldn't this be zero? We start out with 0 outstanding requests, when we begin to handle a request, [we increase `outstanding`](https://github.com/cloudflare/gokeyless/pull/292/files#diff-78f42ba40d0f10b08c73b7e6bb8376f398e249c963cf549e89591d6b6826b9a4R177), when we are done handling it, [we decrease outstanding](https://github.com/cloudflare/gokeyless/pull/292/files#diff-78f42ba40d0f10b08c73b7e6bb8376f398e249c963cf549e89591d6b6826b9a4R535)? Or is that not the case?

server/server.go
```diff
@@ -691,19 +511,43 @@ func (s *Server) spawn(l net.Listener, c net.Conn) {
...
+		limited:     limited,
+		conn:        tconn,
+		listener:    l,
+	}
+	handler.c = sync.NewCond(&handler.mtx)
+
+	for {
+		pkt := new(protocol.Packet)
+		handler.mtx.Lock()
+		for handler.outstanding == 0 {
```
tanyav2, line 532: By similar logic as above, shouldn't this check if `handler.outstanding > maxConnPendingRequests`, i.e, check that we are not trying to process more than `maxConnPendingRequests` at the same time.

server/server.go
wbl: I think I should rename this variable to "tickets" or something.

server/server.go
wbl: Same reply as a above. I start by filling the bucket of tokens to the max.

server/server.go
tanyav2: Or remainingCapacity or capacity

server/server.go
```diff
@@ -301,189 +144,203 @@ type Sealer interface {
...
-type request struct {
-	pkt *protocol.Packet
-	// time just after the request was deserialized from the connection
-	reqBegin time.Time
-	connName string
+type handler struct {
+	name        string
+	outstanding int
+	s           *Server
+	c           *sync.Cond
```
tanyav2, line 151: Instead of using a condition variable, a channel with a capacity of `maxConnPendingRequests` could be used. This way the `outstanding` variable can also be removed since we can just use the channel to indicate the available capacity.

server/server.go
wbl: I've replaced it with a channel

server/server.go
```diff
@@ -691,19 +521,29 @@ func (s *Server) spawn(l net.Listener, c net.Conn) {
...
+	log.Debugf("%s: serving", connStr)
+	handler := &handler{
+		name:     connStr,
+		s:        s,
+		tokens:   make(chan struct{}, s.config.maxConnPendingRequests),
+		limited:  limited,
+		conn:     tconn,
+		listener: l,
+	}
+	for i := 0; i < s.config.maxConnPendingRequests; i++ {
```
tanyav2, line 536: To make this less error prone for code updates, 
```
for len(handler.tokens) < cap(handler.tokens) {
	handler.tokens <- struct{}{}
}
```

server/server.go
```diff
@@ -301,189 +144,213 @@ type Sealer interface {
...
+	respPkt := protocol.Packet{
+		Header: protocol.Header{
+			MajorVers: 0x01,
+			MinorVers: 0x00,
+			Length:    resp.op.Bytes(),
+			ID:        resp.id,
+		},
+		Operation: resp.op,
+	}
+	h.tokens <- struct{}{}
```
tanyav2, line 174: Would be useful to add a comment here or on line 150 where `tokens` is defined, about what it is and how it's used.

server/server.go
```diff
@@ -301,189 +144,213 @@ type Sealer interface {
...
+	_, err := respPkt.WriteTo(h.conn)
+	h.mtx.Unlock()
+	if err != nil {
+		log.Errorf("connection %v: error in writing response %v", h.name, err)
+		h.conn.Close()
+		h.s.mtx.Lock()
+		delete(h.s.listeners[h.listener], h.conn)
+		h.s.mtx.Unlock()
+	}
+	if err != nil {
```
tanyav2, line 185: `err` on line 178 and 185 are the same. I think this one might be for `h.conn.Close` which isn't caught.

wbl: What do we do with an error in close?

tanyav2 (approved):

nickysemenza (approved):

tests/client_test.go
```diff
@@ -374,3 +374,24 @@ func (s *IntegrationTestSuite) TestShutdown() {
...
 	require.NoError(err)
 }
+
+func (s *IntegrationTestSuite) TestManyOps() {
+	require := require.New(s.T())
+	conn, err := s.remote.Dial(s.client)
+	require.NoError(err)
+	defer conn.Close()
+	var wg sync.WaitGroup
+	for i := 0; i < 2000; i++ {
```
tanyav2, line 384: Can you make this 20000 instead? `maxConnPendingRequests` is `1024`, so `2000` isn't big enough to create a situation where we are waiting for a goroutine to become available. 

The situation this would help test would be behavior of the server when there are a significant number of requests that are waiting and there aren't any available goroutines to handle them, and if it can gracefully recover from that state without any large delays.

tanyav2: It might be even better to do something like `i < s.server.maxConnPendingRequests * 20` so these aren't magic numbers, but then the value would have to be exported so maybe just a comment is fine.

wbl: Unfortunately the race detector limits the number of goroutines we can have open at once. We'd also need to alert if there was a "big delay", and at that point we're looking at scheduler details.  This catches the most likely bug of not putting back the token.

tanyav2: Ah didn't even think of the race detector issue, looks like it's a fairly low limit too about 8k. Looks good otherwise!

PR #293, opened 2021-08-10
Restore timeout handling to connection read and write

wbl:


cbroglie (approved):

server/server.go
```diff
@@ -177,6 +178,7 @@ func (h *handler) handle(pkt *protocol.Packet) {
 	}
 	h.tokens <- struct{}{}
 	h.mtx.Lock()
+	h.conn.SetWriteDeadline(time.Now().Add(h.timeout))
```
tanyav2, line 181: This returns an error that should be handled and the mutex should be unlocked and the connection should be closed.

server/server.go
```diff
@@ -195,6 +197,7 @@ func (h *handler) loop(tconn net.Conn) {
 	for {
 		pkt := new(protocol.Packet)
 		<-h.tokens
+		h.conn.SetReadDeadline(time.Now().Add(h.timeout))
```
tanyav2, line 200: Same as with `SetWriteDeadline`, error should be handled and we should break out of loop

server/server.go
tanyav2: Seems like https://github.com/cloudflare/gokeyless/blob/efe77714f3928a83da416f15058252466ed2add9/conn/conn.go#L266-L270 handles the error as well

server/server.go
tanyav2, line 200: Shouldn't this be `tconn` since the next line is `pkt.ReadFrom(tconn)`?

tanyav2 (approved):

lgarofalo (approved):

server/server.go
```diff
@@ -177,27 +187,30 @@ func (h *handler) handle(pkt *protocol.Packet) {
...
 	if err != nil {
-		log.Errorf("connection %v: error in writing response %v", h.name, err)
-		h.conn.Close() // ignoring error: what can we do?
-		h.s.mtx.Lock()
-		delete(h.s.listeners[h.listener], h.conn)
-		h.s.mtx.Unlock()
-		logConnFailure()
+		h.closeWithWritingErr(err)
+	}
+	_, err = respPkt.WriteTo(h.conn)
```
cbroglie, line 195: You shouldn't execute this if you closed the connection due to `SetWriteDeadline` returning an error

tanyav2 (approved):

tanyav2 (commented):
Looks like tests are failing

server/server.go
```diff
@@ -177,27 +187,31 @@ func (h *handler) handle(pkt *protocol.Packet) {
...
+	err := h.conn.SetWriteDeadline(time.Now().Add(h.timeout))
 	if err != nil {
-		log.Errorf("connection %v: error in writing response %v", h.name, err)
-		h.conn.Close() // ignoring error: what can we do?
-		h.s.mtx.Lock()
-		delete(h.s.listeners[h.listener], h.conn)
-		h.s.mtx.Unlock()
-		logConnFailure()
+		h.closeWithWritingErr(err)
+		return // we lose the metrics, but this is an exceptional case.
```
cbroglie, line 194: Definitely an exceptional case, but you could handle it for all cases by moving the metrics logging to a defer:
```go
h.mtx.Lock()
defer h.mtx.Unlock()

defer func() {
	logRequestExecDuration(pkt.Operation.Opcode, start, resp.op.ErrorVal())
	logRequestTotalDuration(pkt.Operation.Opcode, start, resp.op.ErrorVal())
}()

err := h.conn.SetWriteDeadline(time.Now().Add(h.timeout))
if err != nil {
	h.closeWithWritingErr(err)
	return
}

_, err = respPkt.WriteTo(h.conn)
if err != nil {
	h.closeWithWritingErr(err)
	return
}
```

cbroglie (approved):

server/server.go
tanyav2: +1 It also seemed to me that having metrics when an exceptional case is encountered would be even more necessary

PR #295, opened 2021-08-30
Add explicit import as required by go 1.17

wbl:
Go 1.17 makes changes to the ways modules work that require some
additional imports.


nickysemenza (approved):
perhaps we should add `1.17` (maybe there's a way to add `latest` too?) to https://github.com/cloudflare/gokeyless/blob/master/.github/workflows/go.yml#L10

PR #297, opened 2021-09-17
Adjust logging and metrics to match earlier versions

wbl:
We now execute the cleanup function once for connection errors that
result in closure, and measure the total time and execution time
separately. This matches the behavior of versions with the worker
pools.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -157,18 +157,22 @@ type handler struct {
 	listener net.Listener
 	conn     net.Conn
 	timeout  time.Duration
+	closed   bool
 }
 
 func (h *handler) closeWithWritingErr(err error) {
-	log.Errorf("connection %v: error in writing response %v", h.name, err)
-	h.conn.Close() // ignoring error: what can we do?
-	h.s.mtx.Lock()
-	delete(h.s.listeners[h.listener], h.conn)
-	h.s.mtx.Unlock()
-	logConnFailure()
+	if !h.closed {
+		log.Errorf("connection %v: error in writing response %v", h.name, err)
+		h.conn.Close() // ignoring error: what can we do?
+		h.s.mtx.Lock()
+		delete(h.s.listeners[h.listener], h.conn)
+		h.s.mtx.Unlock()
+		logConnFailure()
+		h.closed = true
+	}
 }
 
-func (h *handler) handle(pkt *protocol.Packet) {
+func (h *handler) handle(pkt *protocol.Packet, reqTime time.Time) {
 	var resp response
 	start := time.Now()
 	if h.limited {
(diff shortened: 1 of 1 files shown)


cbroglie (approved):

tanyav2:
Looks like `logKeyLoadDuration` is still missing.

server/server.go
```diff
@@ -190,7 +194,7 @@ func (h *handler) handle(pkt *protocol.Packet) {
 	defer h.mtx.Unlock()
 	defer func() {
 		logRequestExecDuration(pkt.Operation.Opcode, start, resp.op.ErrorVal())
```
tanyav2, line 196: It doesn't make sense why `ExecDuration` should include anything besides the actual execution of the operation. Why not keep it as a part of `makeRespondResponse` etc as it was before:
https://github.com/cloudflare/gokeyless/blob/34973625ac18b695a4735b0e045007b52408ef0d/server/server.go#L320-L333

tanyav2:
While we're adding metrics, one metric I would really like to see is the number of tokens in the channel each time we remove a token and add a token. Also, instead of using an empty struct as a token, would it be too much overhead to instead use time? This way when we remove a token from the channel, we can figure out how long it was in it for?

This would be super helpful for debugging issues we've been having without having to guess what's the best config for things like `maxConnPendingRequests`.

server/server.go
wbl: We're not done until we've sent back the request.

server/server.go
```diff
@@ -311,7 +315,9 @@ func (s *Server) unlimitedDo(pkt *protocol.Packet, connName string) response {
 		return makeRespondResponse(pkt, res)
 
 	case protocol.OpEd25519Sign:
+		loadStart := time.Now()
 		key, err := s.keys.Get(ctx, &pkt.Operation)
+		logKeyLoadDuration(loadStart)
```
tanyav2, line 320: I think this should be logged after the error check. Logging it here can make latencies appear better or worse than they actually are depending on types of errors, for instance timeout errors could skew the duration much higher, and it would be hard to tell.

server/server.go
tanyav2: But that is the purpose of `TotalDuration`, `ExecDuration` shouldn't be concerned with that. The previous version of the code didn't consider the sending bits either. And the actual metric seems to suggest the same:
https://github.com/cloudflare/gokeyless/blob/1c828f26e77404b5111574dbd29f26ef3ba46bd3/server/metrics.go#L21-L27

wbl: Makes sense

server/server.go
wbl: I don't think we should throw out errors. it should be the time for the log key load function to run. If we're getting a bunch of errors other metrics will go high like the error count. We don't have a great way to granularize metrics because Prometheus metrics are expensive.

tanyav2 (approved):

server/server.go
```diff
@@ -189,8 +194,7 @@ func (h *handler) handle(pkt *protocol.Packet) {
 	h.mtx.Lock()
 	defer h.mtx.Unlock()
 	defer func() {
```
tanyav2, line 196: nit: `defer logRequestTotalDuration(pkt.Operation.Opcode, reqTime, resp.op.ErrorVal())`

PR #299, opened 2021-10-27
Better timeout handling

wbl:
When a read timeout happens, instead of going away, we tear down the
connection after letting any pending requests finish.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -24,6 +24,7 @@ import (
 	"github.com/opentracing/opentracing-go/ext"
 
 	"golang.org/x/crypto/ed25519"
+	"golang.org/x/sync/semaphore"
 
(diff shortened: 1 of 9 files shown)


mihirjham (approved):

server/server.go
```diff
@@ -207,20 +214,41 @@ func (h *handler) handle(pkt *protocol.Packet, reqTime time.Time) {
...
 		_, err = pkt.ReadFrom(h.conn)
 		if err != nil {
-			return err
+			h.tokens.Release(1)
+			break
 		}
 		go h.handle(pkt, time.Now())
-
 	}
+	neterr, ok := err.(net.Error)
```
nickysemenza, line 236: I think `errors.As` might be a bit more futureproof here

nickysemenza (approved):

PR #300, opened 2021-11-06
Enable concurrent outstanding RPC requests

wbl:


wbl:
i actually need to redo this with more complexity to handle timeouts right

wbl:
That better fix is now ready for a look

conn/conn.go
```diff
@@ -219,15 +219,14 @@ func (c *Conn) timeoutRequest(id uint32, timeout time.Duration) {
...
 	}
-	place <- &result{err: fmt.Errorf("operation timed out")}
+	place <- &result{err: ErrTimeout}
 }
 
-// DoOperation executes an entire keyless operation, returning its result.
-func (c *Conn) DoOperation(ctx context.Context, op protocol.Operation) (*protocol.Operation, error) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, "Conn.DoOperation")
+// sendOp sends operation, returning a channel to wait for results
+func (c *Conn) sendOp(ctx context.Context, op protocol.Operation) (chan *result, error) {
```
tanyav2, line 226: Does this need to be exported? On the gokeyless-internal front, we can't use `DoOperation` directly anymore, since that still waits synchronously for the response.

wbl: I don't think it does: we have a goroutine per request model and hook through the RPC support here. That didn't work for net/rpc because net/rpc doesn't do well with packetized protocols.

nickysemenza (approved):

PR #305, opened 2022-07-13
downgrade log level of noisy EOF

nickysemenza:
`[ERROR] closing connection connection 127.0.0.1:21076: read error EOF` to debug level with a varying port number is common if the other end closes the conn (in which case we don't need to gracefully do anything).


tanyav2 (approved):
Nice fix, this always pollutes the logs!

PR #308, opened 2022-10-21
Fix protocol reference

jyn514:
According to https://github.com/cloudflare/keyless/blob/c84cb94f78b3b030c2637316296bd09c6e438331/kssl.h#L89, 0xFF is an error opcode, not anything to do with RSA keys.


nickysemenza (approved):
for the record, the repo you linked is the outdated c version, https://github.com/cloudflare/gokeyless/blob/master/protocol/protocol.go#L115 is the modern protocol reference (but in terms of this fix, they both agree that `0xFF` is error)

PR #311, opened 2023-02-14
Add function to access remote connection pool

lgarofalo:


maxnystrom (approved):

client/remote.go
```diff
@@ -137,6 +137,11 @@ func healthchecker(c *Conn) {
 	}
 }
 
+// GetConnPool allows other types of remotes to interact with the connection pool
+func GetConnPool() *connPoolType {
```
nickysemenza, line 141: i agree with the lint here, but i suppose practically speaking this return value is just being passed back into this package so it's probably not an actual issue

nickysemenza (approved):

client/remote.go
lgarofalo: Yea, not an issue but it's a reasonable lint

PR #314, opened 2023-03-22
add goreleaser snapshot and release action, enable pkcs11 in built binaries

nickysemenza:
now if you go to https://github.com/cloudflare/gokeyless/actions/runs/4494483693 (test run from every commit) you can download a binary for linux + macOS

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,23 @@
+name: goreleaser
+
+on:
+  push:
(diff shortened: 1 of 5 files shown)


vasilzhigilei:
Huh, interesting how it removed "SSL Team" from reviewers after I added my review. I wonder if that's just a UI thing or if it really did replace SSL Team with me?

PR #350, opened 2023-05-16
update services to expect binary in usr/bin

### LINKED ISSUE #328: update packaged systemd service to start on boot

nickysemenza:
This seems to be more correct than user/local/bin, and goreleaser's defaults corroborate that.

Resolves #328, as this behavior is the systemd default Adds on to #320 as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ VERSION := $(shell git describe --tags --abbrev=0 | tr -d '[:alpha:]')
 LDFLAGS := "-X main.version=$(VERSION)"
 
 DESTDIR                      := build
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -7,7 +7,7 @@ VERSION := $(shell git describe --tags --abbrev=0 | tr -d '[:alpha:]')
 LDFLAGS := "-X main.version=$(VERSION)"
 
 DESTDIR                      := build
-PREFIX                       := usr/local
+PREFIX                       := usr
```
mitalirawat, line 10: 👍

mitalirawat (approved):

PR #366, opened 2023-06-16
Testing retry and timeout for signing ops

lgarofalo:


server/server.go
```diff
@@ -451,16 +456,52 @@ func (s *Server) unlimitedDo(pkt *protocol.Packet, connName string) response {
...
 
 	return makeRespondResponse(pkt, sig)
 }
 
+type signWithTimeoutWrapper struct {
+	sig   []byte
+	error error
+}
+
+func signWithTimeout(ctx context.Context, ch chan signWithTimeoutWrapper, key crypto.Signer, rand io.Reader, digest []byte, opts crypto.SignerOpts) {
```
nickysemenza, line 500: Maybe a future refactor, it looks like we should replace `crypto.Signer` with something that has `context.Context` in the signature, that way the sign operation can actually be cancelled after the timeout, not just discarded (but kept running)

nickysemenza (approved):

server/server.go
lgarofalo: Yea, I didn't see a way to do this without modifying the implementation of the signer inside the vendored dependencies, but if there was a way to do this outside I'd love to change it.

PR #389, opened 2023-08-24
Adjust reporting of connection failure metric

lgarofalo:


server/server.go
```diff
@@ -166,21 +166,23 @@ type handler struct {
...
-func (h *handler) close() {
+func (h *handler) close(err bool) {
 	if !h.closed {
 		h.conn.Close() // ignoring error: what can we do?
 		h.s.mtx.Lock()
 		delete(h.s.listeners[h.listener], h.conn)
 		h.s.mtx.Unlock()
-		logConnFailure()
 		h.closed = true
+		if err {
```
nickysemenza, line 176: super minor, should this func accept `err error` and this check be 
```suggestion
		if err != nil {
```
doesn't really matter though i guess

lgarofalo: Yea I like that, I'll update.

PR #391, opened 2023-09-01
Add docker image build configuration

lgarofalo:
I will update this with some documentation about using the containerized gokeyless, and I believe I can add some integration test since gokeyless doesn't need to phone home unless to get new certs, but we already have dev certs configured in the repo.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,3 +21,47 @@ jobs:
       - run: make release-github
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    strategy:
+      matrix:
+        include:
+          # github container registry
+          - registry: "ghcr.io"
+            username: ${{ github.actor }}
+            password_secret: GITHUB_TOKEN
+            image: ghcr.io/cloudflare/gokeyless
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Log in to the Docker hub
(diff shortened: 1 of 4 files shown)


docker-compose.yaml
lgarofalo, line 1: This is mostly for demonstration, the customer would need to configure their own containerization setup, just an example.

docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
...
+    build:
+      context: .
+      dockerfile: Dockerfile
+    volumes:
+      - ./tests/testdata/:/gokeyless/config
+    environment:
+      # - KEYLESS_HOSTNAME=keyserver.keyless.com
+      # - KEYLESS_ZONE_ID=5c7004f5221ba2a24c998bd609244a39
+      # - KEYLESS_ORIGIN_CA_API_KEY=ORIGIN_CA_API_KEY_HERE
+      - KEYLESS_AUTH_CERT=/gokeyless/config/server.pem
```
nickysemenza, line 15: I might be missing something, is the code capable of reading this env vars?

.github/workflows/release.yml
```diff
@@ -21,3 +21,47 @@ jobs:
       - run: make release-github
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  build-and-push-image:
```
nickysemenza, line 24: do you think it's worth adding this to a new workflow file that runs on every commit/push? Or is just tagged images sufficient?

nickysemenza (approved):

docker-compose.yaml
lgarofalo: Yeap, it's hooked in around this https://github.com/cloudflare/gokeyless/blob/master/cmd/gokeyless/gokeyless.go#L157

.github/workflows/release.yml
lgarofalo: True, I think only on tagged is sufficient since its the same build process that exists currently, I might cover this if I were to add tests around the containerized gokeyless to the integration tests.

docker-compose.yaml
nickysemenza: ah, the prefix killing searchability!

PR #510, opened 2025-03-06
Adds ComplianceRegion to Op protocol

mitch292:


vasilzhigilei (approved):

mitalirawat (approved):

protocol/protocol.go
```diff
@@ -673,6 +700,11 @@ func (o *Operation) UnmarshalBinary(body []byte) error {
 			o.JaegerSpan = data
 		case TagReqContext:
 			o.ReqContext = data
+		case TagComplianceRegion:
+			if len(data) != 1 {
+				return fmt.Errorf("invalid ComplianceRegion: %s", data)
+			}
+			o.ComplianceRegion = ComplianceRegion(data[0])
```
schien, line 707: sorry for the confusion in my spec, the value we'll be sending in this TLV will be a "fedramp_high" UTF-8 string instead.

mitch292: This was discussed further and we're going to do the byte constant as the value for compliance region

PR #512, opened 2025-03-11
Resolves some linting errors flagged in GitHub actions

mitch292:
Fixes a few of the linting errors that GithubActions has been flagging.


vasilzhigilei (approved):

mitch292:
ah I guess there is a limit to the amount of linting errors it will show, because now new ones show. I'm going to actually just close this PR and create an issue instead. We could run the linter locally and fix them in one go at a different point.

PR #522, opened 2025-04-15
Do not add Compliance Region to the packet if it's a 0 value

mitch292:
We shouldn't add the 0 value by default to the packet, if included it could be a breaking change for some clients.

### DIFF
--- a/protocol/protocol.go
+++ b/protocol/protocol.go
@@ -527,7 +527,9 @@ func (o *Operation) Bytes() uint16 {
 		add(tlvLen(len(o.ReqContext)))
 	}
 	// compliance region
-	add(tlvLen(1))
+	if o.ComplianceRegion != 0 {
+		add(tlvLen(1))
+	}
 	if int(length)+headerSize < paddedLength {
 		// TODO: Are we sure that's the right behavior?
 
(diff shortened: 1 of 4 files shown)


lgarofalo (approved):

protocol/protocol.go
```diff
@@ -602,7 +604,9 @@ func (o *Operation) MarshalBinary() ([]byte, error) {
 		b = append(b, tlvBytes(TagReqContext, o.ReqContext)...)
 	}
 
-	b = append(b, tlvBytes(TagComplianceRegion, []byte{byte(o.ComplianceRegion)})...)
+	if o.ComplianceRegion != 0 {
```
mitch292, line 607: When we are returning packets, the Op is built here -https://github.com/cloudflare/gokeyless/blob/master/protocol/conn.go#L41 - it will be the 0 value for op 100% of the time. With the current implementation that tag would've never been accurate for FRH. There doesn't seem to be value in returning the tlv back to the consumer anyway.

mitch292: We should at least add a comment here that including compliance region is a breaking change for at least 1 client. It's a little fragile as is. Maybe we add a test as well that will fail when we return compliance region.

mitalirawat (approved):

PR #563, opened 2026-06-30
Support Cloudflare API token for server cert initialization

lgarofalo:
The server-cert enrollment flow only supported the deprecated Origin CA Service Key (X-Auth-User-Service-Key). Cloudflare now recommends API tokens (Authorization: Bearer) scoped Zone > SSL and Certificates > Edit for the Origin CA /certificates endpoint.

Route the init request through the vendored cloudflare-go client so it selects the correct auth scheme automatically, preferring an API token and falling back to the (now deprecated) service key. The request still uses Raw() so the keyless-specific request_type and zone_id body fields are preserved (the typed CreateOriginCACertificate helper does not expose them).


lgarofalo:
Verified end to end as well with live zone.

mitch292 (approved):
lgtm

